PR #6, opened 2017-11-23
widen integer types

SimonDanisch:
I was surprised that indexing with e.g. UInt32 works, but it turned out that it just routes me through some quite involved base fallback for that case, yielding pretty bad code (at least for my transpiler - might still compile down to only a couple of llvm ir instructions). 
Anyways, there is no reason why not just use `Integer` everywhere!

### DIFF
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -6,12 +6,12 @@ export Zeros, Ones, Fill, Eye
 
 abstract type AbstractFill{T, N} <: AbstractArray{T, N} end
 
-@inline function getindex(F::AbstractFill, k::Int)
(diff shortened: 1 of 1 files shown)


SimonDanisch:
Ohh I see, I activated appveyor, but it seems like you deleted the appveyor script?

dlfivefifty:
I don't think there ever was an appveyor script.

SimonDanisch:
Oh, but `PkgDev.generate` should generate that

dlfivefifty:
Ah, I never ran PkgDev.generate….

SimonDanisch:
Oh, crazy! :)

PR #14, opened 2018-04-26
Implement multiplication by zero identities

cstjean:
Defining so many methods doesn't feel great. Does anyone know a better way?

If this is good, I'll do `Eye` identities in the next PR.

### DIFF
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -245,4 +245,28 @@ convert(::Type{AbstractSparseArray{Tv,Ti}}, Z::Eye{T}) where {T,Tv,Ti} =
 convert(::Type{AbstractSparseArray{Tv,Ti,2}}, Z::Eye{T}) where {T,Tv,Ti} =
     convert(SparseMatrixCSC{Tv,Ti}, Z)
 
+## Algebraic identities
+dim_mismatch() =
+    throw(DimensionMismatch("Incompatible matrix multiplication dimensions"))
+
+for a in 1:2, b in 1:2
+    # We cannot use Unions{AbstractArray, AbstractVector}, because Julia's dispatch
+    # then picks the generic *(::AbstractArray, ::AbstractArray) method.
+    # There might be a way to use promotion rules to avoid adding so many methods,
+    # but I don't think so.
+    @eval function Base.:*(z::Zeros{T, $a}, v::AbstractArray{U, $b}) where {T, U}
+        size(z, 2) ≠ size(v, 1) && dim_mismatch()
+        Zeros{promote_type(T, U)}(size(z, 1), size(v, 2))
+    end
+    @eval function Base.:*(v::AbstractArray{U, $b}, z::Zeros{T, $a}) where {T, U}
+        size(v, 2) ≠ size(z, 1) && dim_mismatch()
+        Zeros{promote_type(T, U)}(size(v, 1), size(z, 2))
+    end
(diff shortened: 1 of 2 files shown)


src/FillArrays.jl
```diff
@@ -245,4 +245,28 @@ convert(::Type{AbstractSparseArray{Tv,Ti}}, Z::Eye{T}) where {T,Tv,Ti} =
...
+    throw(DimensionMismatch("Incompatible matrix multiplication dimensions"))
+
+for a in 1:2, b in 1:2
+    # We cannot use Unions{AbstractArray, AbstractVector}, because Julia's dispatch
+    # then picks the generic *(::AbstractArray, ::AbstractArray) method.
+    # There might be a way to use promotion rules to avoid adding so many methods,
+    # but I don't think so.
+    @eval function Base.:*(z::Zeros{T, $a}, v::AbstractArray{U, $b}) where {T, U}
+        size(z, 2) ≠ size(v, 1) && dim_mismatch()
+        Zeros{promote_type(T, U)}(size(z, 1), size(v, 2))
```
dlfivefifty: This is not correct for matrix * vector, which should return a vector

dlfivefifty:
It looks a bit overengineered for my tastes (sometimes it’s better for readability to repeat code with copy-and-paste), but I’ll merge it.

cstjean:
Fair enough. I plan to do a few such PRs, so we may as well agree on a particular style. Would you rather have nine separate methods? Or perhaps I could write the loops more explicitly similar to [this](https://github.com/cstjean/FillArrays.jl/blob/84857e5eecc23cd1b5ea798167415dd378bc8851/src/FillArrays.jl#L62-L94)?

Come to think of it, writing separate methods like

```julia
Base.:*(a::Zeros{<:Any, 2}, b::AbstractMatrix) = mult_zeros(a,b)
Base.:*(a::Zeros{<:Any, 1}, b::AbstractMatrix) = mult_zeros(a,b)
...
```

with the dimension check in `mult_zeros` wouldn't be so bad.

dlfivefifty:
I'd suggest following the style of `UniformScaling` in Base/StdLib:  https://github.com/JuliaLang/julia/blob/ef76d9cd19ecb4fe47d58070d3116357283cb3a5/stdlib/LinearAlgebra/src/uniformscaling.jl#L83

cstjean:
```julia
function mult_zeros(a, b::AbstractMatrix)
    size(a, 2) ≠ size(b, 1) &&
        throw(DimensionMismatch("Incompatible matrix multiplication dimensions"))
    return Zeros{promote_type(eltype(a), eltype(b))}(size(a, 1), size(b, 2))
end
function mult_zeros(a, b::AbstractVector)
    size(a, 2) ≠ size(b, 1) &&
        throw(DimensionMismatch("Incompatible matrix multiplication dimensions"))
    return Zeros{promote_type(eltype(a), eltype(b))}(size(a, 1))
end

Base.:*(a::Zeros, b::AbstractMatrix) = mult_zeros(a, b)
Base.:*(a::AbstractMatrix, b::Zeros) = mult_zeros(a, b)
Base.:*(a::Zeros, b::AbstractVector) = mult_zeros(a, b)
Base.:*(a::AbstractVector, b::Zeros) = mult_zeros(a, b)
Base.:*(a::Zeros, b::Zeros) = mult_zeros(a, b)
```
Better?

PR #16, opened 2018-05-07
Addition with Zeros

willtebbutt:
Adds support for efficient addition of Zeros and other matrices where possible. Covers addition of

- Zeros with Zeros,
- Zeros with AbstractArray
- AbstractArray with Zeros

Care has been taken to return a matrix with the correct `eltype`.

### DIFF
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -265,6 +265,11 @@ Base.:*(a::ZerosVecOrMat, b::AbstractVector) = mult_zeros(a, b)
 Base.:*(a::AbstractVector, b::ZerosVecOrMat) = mult_zeros(a, b)
 Base.:*(a::ZerosVecOrMat, b::ZerosVecOrMat) = mult_zeros(a, b)
 
+
+# Unary = / -.
+Base.:+(a::Zeros) = a
+Base.:-(a::Zeros) = a
+
 # eltypes are the same, so result of promotion is guaranteed to be the same.
 function Base.:+(a::Zeros{T, N}, b::AbstractArray{T, N}) where {T, N}
     size(a) ≠ size(b) && throw(DimensionMismatch("Incompatible matrix add dimensions."))
@@ -285,9 +290,6 @@ function Base.:+(a::Zeros{T, N}, b::Zeros{V, N}) where {T, V, N}
     size(a) ≠ size(b) && throw(DimensionMismatch("Incompatible matrix add dimensions."))
     return Zeros{promote_type(T, V)}(size(a)...)
 end
-
-# Unary negation.
-Base.:-(a::Zeros) = a
 
 # Subtraction can just re-use addition.
 Base.:-(a::Zeros{T, N}, b::AbstractArray{V, N}) where {T, V, N} = -(a + b)
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -281,10 +281,22 @@ function Base.:+(a::Zeros{T, N}, b::Zeros{V, N}) where {T, V, N}
     size(a) ≠ size(b) && throw(DimensionMismatch("dimensions must match."))
     return Zeros{promote_type(T, V)}(size(a)...)
 end
+function Base.:+(a::Zeros{T, N}, b::Range) where {T, N}
+    size(a) ≠ size(b) && throw(DimensionMismatch("dimensions must match."))
+    Tout = promote_type(T, eltype(b))
+    return convert(Tout, first(b)):convert(Tout, step(b)):convert(Tout, last(b))
+end
+function Base.:+(a::Zeros{T, N}, b::UnitRange) where {T, N}
+    size(a) ≠ size(b) && throw(DimensionMismatch("dimensions must match."))
+    Tout = promote_type(T, eltype(b))
+    return convert(Tout, first(b)):convert(Tout, last(b))
+end
+
+Base.:+(a::Range, b::Zeros) = b + a
 
 function Base.:-(a::Zeros{T, N}, b::AbstractArray{V, N}) where {T, V, N}
     size(a) ≠ size(b) && throw(DimensionMismatch("dimensions must match."))
-    return -b
+    return -b + a
 end
 Base.:-(a::AbstractArray{T, N}, b::Zeros{V, N}) where {T, V, N} = a + b
 Base.:-(a::Zeros{T, N}, b::Zeros{V, N}) where {T, V, N} = -(a + b)
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -265,4 +265,25 @@ Base.:*(a::ZerosVecOrMat, b::AbstractVector) = mult_zeros(a, b)
 Base.:*(a::AbstractVector, b::ZerosVecOrMat) = mult_zeros(a, b)
 Base.:*(a::ZerosVecOrMat, b::ZerosVecOrMat) = mult_zeros(a, b)
 
+# eltypes are the same, so result of promotion is guaranteed to be the same.
+function Base.:+(a::Zeros{T, N}, b::AbstractArray{T, N}) where {T, N}
+    size(a) ≠ size(b) && throw(DimensionMismatch("Incompatible matrix add dimensions."))
+    return b
+end
+Base.:+(a::AbstractArray{T, N}, b::Zeros{T, N}) where {T, N} = b + a
+
+# Mismatched eltypes, so have to be careful to select the appropriate eltype.
+function Base.:+(a::Zeros{T, N}, b::AbstractArray{V, N}) where {T, V, N}
+    size(a) ≠ size(b) && throw(DimensionMismatch("Incompatible matrix add dimensions."))
+    return promote_type(T, V) == V ? b : copy!(similar(b, promote_type(T, V)), b)
+end
+function Base.:+(a::AbstractArray{T, N}, b::Zeros{V, N}) where {T, V, N}
+    size(a) ≠ size(b) && throw(DimensionMismatch("Incompatible matrix add dimensions."))
+    return promote_type(T, V) == T ? a : copy!(similar(a, promote_type(T, V)), a)
+end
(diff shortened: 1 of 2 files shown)


dlfivefifty:
Can you also add subtraction?

willtebbutt:
Thanks for the suggestion. Added subtraction as requested.

dlfivefifty:
👍 can you add tests for the 1 argument versions +(::Zeros) and -(::Zeros)?

src/FillArrays.jl
```diff
@@ -265,4 +265,35 @@ Base.:*(a::ZerosVecOrMat, b::AbstractVector) = mult_zeros(a, b)
 Base.:*(a::AbstractVector, b::ZerosVecOrMat) = mult_zeros(a, b)
 Base.:*(a::ZerosVecOrMat, b::ZerosVecOrMat) = mult_zeros(a, b)
 
+
+# Unary = / -.
+Base.:+(a::Zeros) = a
+Base.:-(a::Zeros) = a
+
+# eltypes are the same, so result of promotion is guaranteed to be the same.
+function Base.:+(a::Zeros{T, N}, b::AbstractArray{T, N}) where {T, N}
```
cstjean: Is this method helpful? I would have thought that the `promote_type(T, V) == T ? a : ...` case below would compile down to the same thing.

willtebbutt: Good point, I hadn't thought about that. The following give identical results:
```julia
@code_warntype Zeros{Float64}(5, 4) + randn(5, 4)
@code_warntype Zeros{Float32}(5, 4) + randn(5, 4)
```
which suggests that you are correct, but I'm not entirely confident I've not missed something subtle. If you think this confirms what you're saying, I'll make the appropriate changes.

cstjean: I'm not sure either, but it seems best to start with fewer methods, then add the extra method if an exception is found.

src/FillArrays.jl
```diff
@@ -265,4 +265,35 @@ Base.:*(a::ZerosVecOrMat, b::AbstractVector) = mult_zeros(a, b)
...
+    size(a) ≠ size(b) && throw(DimensionMismatch("Incompatible matrix add dimensions."))
+    return b
+end
+Base.:+(a::AbstractArray{T, N}, b::Zeros{T, N}) where {T, N} = b + a
+
+# Mismatched eltypes, so have to be careful to select the appropriate eltype.
+function Base.:+(a::Zeros{T, N}, b::AbstractArray{V, N}) where {T, V, N}
+    size(a) ≠ size(b) && throw(DimensionMismatch("Incompatible matrix add dimensions."))
+    return promote_type(T, V) == V ? b : copy!(similar(b, promote_type(T, V)), b)
```
dlfivefifty: This seems like a bad idea: one expects `0 .+ b` to make a copy of `b` and so not making a copy may cause inconsistent bugs in user code.

If a user really doesn't want a copy, they should use `BLAS.axpy!`.

src/FillArrays.jl
```diff
@@ -265,4 +265,35 @@ Base.:*(a::ZerosVecOrMat, b::AbstractVector) = mult_zeros(a, b)
...
+    size(a) ≠ size(b) && throw(DimensionMismatch("Incompatible matrix add dimensions."))
+    return promote_type(T, V) == T ? a : copy!(similar(a, promote_type(T, V)), a)
+end
+function Base.:+(a::Zeros{T, N}, b::Zeros{V, N}) where {T, V, N}
+    size(a) ≠ size(b) && throw(DimensionMismatch("Incompatible matrix add dimensions."))
+    return Zeros{promote_type(T, V)}(size(a)...)
+end
+
+# Subtraction can just re-use addition.
+Base.:-(a::Zeros{T, N}, b::AbstractArray{V, N}) where {T, V, N} = -(a + b)
```
dlfivefifty: If `a+b` makes a copy, then this is wasteful. Just do this:
```julia
function Base.:-(a::Zeros{T, N}, b::AbstractArray{V, N}) where {T, V, N}
size(a) ≠ size(b) && throw(DimensionMismatch("Incompatible matrix add dimensions."))
-b
end

File: src/FillArrays.jl [hunk 4]
```diff
@@ -265,4 +265,35 @@ Base.:*(a::ZerosVecOrMat, b::AbstractVector) = mult_zeros(a, b)
...
+end
+Base.:+(a::AbstractArray{T, N}, b::Zeros{T, N}) where {T, N} = b + a
+
+# Mismatched eltypes, so have to be careful to select the appropriate eltype.
+function Base.:+(a::Zeros{T, N}, b::AbstractArray{V, N}) where {T, V, N}
+    size(a) ≠ size(b) && throw(DimensionMismatch("Incompatible matrix add dimensions."))
+    return promote_type(T, V) == V ? b : copy!(similar(b, promote_type(T, V)), b)
+end
+function Base.:+(a::AbstractArray{T, N}, b::Zeros{V, N}) where {T, V, N}
+    size(a) ≠ size(b) && throw(DimensionMismatch("Incompatible matrix add dimensions."))
```
dlfivefifty: This error message should match the standard one:
```julia
julia> rand(5)+rand(3)
ERROR: DimensionMismatch("dimensions must match")
Stacktrace:
 [1] promote_shape(::Tuple{Base.OneTo{Int64}}, ::Tuple{Base.OneTo{Int64}}) at ./indices.jl:79
 [2] +(::Array{Float64,1}, ::Array{Float64,1}) at ./arraymath.jl:38
```

src/FillArrays.jl
cstjean: > one expects 0 .+ b to make a copy of b and so not making a copy

Code that assumes a copy is code that will mutate `b`, but that code is already broken by commonly-used immutable data structures, like `b::UnitRange`, `b::UniformScaling`, `b::SVector`, and `b::Zeros`. 

I get what you mean in theory, but I'd be curious to see actual code where this is an issue. Is there precedent in Base one way or the other?

dlfivefifty: If a type is immutable, then there is no difference between a copy and not. The issue here is that someone might do something like this:
```julia
function foo(A, B)
       C = A + B
       C[1] = 2
       C
       end
```
This function is broken for `foo(::Zeros, ::Vector)` using the current definition of `+`, as it will modify the input variable `B`.

cstjean: What I mean is that `foo` is already broken for `foo(SVector(1,2), SVector(3,4))`, or `foo(1:2, 3:4)` because `setindex!` is not defined for those data structures.

My point is that if you're writing generic code (the kind that accepts exotic array types like `Zeros`), you should not rely on mutability, and so it does not matter if `b + 0` returns a copy or not.

dlfivefifty: "Broken" by throwing an error is acceptable. Broken by silently mutating the input is not. 

In every other case, `+(A::AbstractVector, B::Vector)` will not return `B`, and there is no way a user will know or expect  it to return `B` in the special case of `A` being a `Zeros`.

cstjean: Coming back to the problem at hand, @willtebbutt  `copy!(similar(b, promote_type(T, V)), b)` turns a `SVector` into an `MVector` and an `UnitRange` into a `Vector`. Maybe those should be special-cased?  Or perhaps the `copy!` should be turned into `map(x->convert(promote_type(T,V), x), b)`? That works for `SVector`, but not ranges.

dlfivefifty: If it actually makes a difference in practice (which seems unlikely since `copy` is pretty fast), we could add a function `lazy_add`, and a macro (e.g. `@lazy_arithmetic`) that changes calls to `+` to calls to `lazy_add`.

cstjean: It's not the copy I'm worried about, it's the memory allocation. But for my use case, I'm likely to use `SMatrix`, so it's moot.

I'm working on Kalman filters, and I liked the idea that if my fully-general code is passed `transition_matrix = Eye(3)`, then its performance will be the same as if I had written a specialized version of the code assuming an identity transition. It bugs me that I'll pay the cost of a `copy`, but I agree that it's most likely negligible compared to the other operations.

dlfivefifty: My suggestion to write code that is performant would be to write things like:
```julia
A .= B .+ C
```
Then we could add an override like (it'll be more complicated when done properly)
```julia
broadcast!(::typeof(+), dest::AbstractVector, A::Zeros, B::AbstractVector) = dest === B ? B : (dest .= B)
```

dlfivefifty:
@cstjean FYI some supporting evidence that one should make a copy:
```julia
julia> X = randn(5); I*X === X
false
```

@willtebbutt Can we add the following tests:
```julia
X = randn(5)
@test Zeros(5) + X !=== X
@test Zeros{Int}(5) + (1:5) === (1:5)
@test Zeros(5) + (1:5) === (1.0:1.0:5.0)
```

Some supporting evidence that we should return a range:
```julia
julia> I*(1:5)
1:1:5
```

willtebbutt:
Regarding the range issue, the following edge case worries me:
```julia
julia> a, b, c=  1, 2^53, 2^54

julia> collect(a:b:c)
2-element Array{Int64,1}:
                1
 9007199254740993

julia> collect(Float64(a):Float64(b):Float64(c))
3-element Array{Float64,1}:
 1.0       
 9.0072e15 
 1.80144e16
```
This happens because the range of integers for which `Float64`s are exact is smaller than that for `Ints`.
There are a few of options:
1. Ignore it and hope it never happens (probably bad)
2. Don't use a sparse representation at all when precision can't be guaranteed based on type information alone (I believe this is the current implementation).
3. Determine dynamically when doing this makes sense.

dlfivefifty:
I think the answer is "Ignore it and hope it never happens (probably bad)", as this is what is done with overflow.

Note ranges support `BigInt` so if you are using large ints near overflow, it's your fault for not switching to `BigInt`.

willtebbutt:
Looks like everything is good now, although it's possible I've missed something in the `AbstractRange` type hierarchy.

PR #24, opened 2018-08-17
Resolve some AdjOrTrans * Zeros{T, 1} ambiguities

willtebbutt:
Fixes ambiguities arising from the following:
```julia
using FillArrays
randn(5, 4)' * Zeros(5)
randn(5)' * Zeros(5)
```
and for the `transpose` equivalents.


willtebbutt:
Thanks for the review. Agreed, it should be 0.7 not 1.0.

PR #27, opened 2018-09-04
Fix bool indexing

### LINKED ISSUE #26: Bad indexing in latest version
With FillArrays 0.2.0:

```julia
julia> using FillArrays

julia> Fill(10, 5)[[true, true, false, true, false]]
5-element Fill{Int64,1,Tuple{Int64}}:
 10
 10
 10
 10
 10

julia> fill(10, 5)[[true, true, false, true, false]]
3-element Array{Int64,1}:
 10
 10
 10
```

cstjean:
Fix #26, the easy way, using your proposed solution.

### DIFF
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -73,12 +73,14 @@ function getindex(F::Fill, kj::Vararg{AbstractVector{II},N}) where {II<:Integer,
     Fill(getindex_value(F), length.(kj))
 end
 
-function getindex(A::Fill, kr::AbstractArray{Bool})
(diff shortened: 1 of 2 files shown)


dlfivefifty:
Can you add tests for the `DimensionMismatch`?

cstjean:
Done.

dlfivefifty:
It looks like there are two missing tests for 💯 code coverage: `Fill` with `AbstractArray{Bool}` and `Ones`/`Zeros` with `AbstractVector{Bool}`

PR #41, opened 2018-11-08
Conversion to Fill

### LINKED ISSUE #38: Conversion from array
What would you say to supporting `convert(Fill, [2.0,2.0,2.0])`, and throwing an error if it's not uniformly the same value?

cstjean:
I was a bit discouraged by the endless number of conversion methods and potential ambiguities, so this is fairly minimal. Should we define custom exception types? Fix #38

### DIFF
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -84,6 +84,7 @@ function unique_value(arr::AbstractArray)
     end
     return val
 end
+unique_value(f::AbstractFill) = getindex_value(f)
 convert(::Type{Fill}, arr::AbstractArray{T}) where T = Fill{T}(unique_value(arr), axes(arr))
 convert(::Type{Fill{T}}, arr::AbstractArray) where T = Fill{T}(unique_value(arr), axes(arr))
 convert(::Type{T}, F::T) where T<:Fill = F   # ambiguity fix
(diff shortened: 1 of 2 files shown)


src/FillArrays.jl
```diff
@@ -73,6 +73,22 @@ convert(::Type{AbstractFill}, F::AbstractFill) = F
...
+    for x in arr
+        if x !== val
+            error("Input array contains both $x and $val. Cannot convert to Fill")
+        end
+    end
+    return val
+end
+convert(::Type{Fill}, arr::AbstractArray{T}) where T = Fill{T}(unique_value(arr), axes(arr))
+convert(::Type{Fill{T}}, arr::AbstractArray) where T = Fill{T}(unique_value(arr), axes(arr))
+convert(::Type{T}, F::T) where T<:Fill = F   # ambiguity fix
```
dlfivefifty: Add the following:
```julia
convert(::Type{Fill{T}}, F::AbstractFill) = Fill{T}(getindex_value(F), axes(F))
convert(::Type{Fill}, F::AbstractFill{T}) where T = Fill{T}(getindex_value(F), axes(F))
```

cstjean: Or just `unique_value(f::AbstractFill) = getindex_value(f)`?

cstjean: Done.

PR #43, opened 2018-11-22
RFC: reintroduce eye(n) (and eye(n,m)), deprecate Eye constructor

iamed2:
~~The functionality is pretty useful but the code encouraged by the deprecation was pretty verbose. 
This PR changes the deprecation to a function of `Eye`.~~

~~While pushing this I realized that adding the method to `Diagonal{...}` is not a good idea, so I will add the `eye` function here as well.~~

Edit: Now I've brought back `eye`. The constructors were type piracy and it wasn't so bad with `Eye` but `SubEye` was too much. 

Some of the BigInt-related tests failed because Julia does not currently handle BigInt dimensions appropriately. 
I will be submitting a PR for that to Julia itself.

### DIFF
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -199,9 +199,12 @@ rank(F::Ones) = 1
 
 
 const Eye{T, Axes} = Diagonal{T, Ones{T,1,Tuple{Axes}}}
+const SubEye{T, EyeAxes, Axes<:NTuple{2, AbstractUnitRange}} = SubArray{T,2,Eye{T,EyeAxes},Axes,false}
 
 Eye{T}(n::Integer) where T = Diagonal(Ones{T}(n))
 Eye(n::Integer) = Diagonal(Ones(n))
+Eye(n::Integer, m::Integer) = view(Eye(max(n,m)), Base.OneTo(n), Base.OneTo(m))
+Eye{T}(n::Integer, m::Integer) where T = view(Eye{T}(max(n,m)), Base.OneTo(n), Base.OneTo(m))
 
 function iterate(iter::Eye, istate = (1, 1))
     (i::Int, j::Int) = istate
@@ -217,8 +220,6 @@ for f in (:permutedims, :triu, :triu!, :tril, :tril!, :inv)
     @eval ($f)(IM::Eye) = IM
 end
 
-@deprecate Eye(n::Integer, m::Integer) view(Eye(max(n,m)), 1:n, 1:m)
-@deprecate Eye{T}(n::Integer, m::Integer) where T view(Eye{T}(max(n,m)), 1:n, 1:m)
 @deprecate Eye{T}(sz::Tuple{Vararg{Integer,2}}) where T Eye{T}(sz...)
 @deprecate Eye(sz::Tuple{Vararg{Integer,2}}) Eye{Float64}(sz...)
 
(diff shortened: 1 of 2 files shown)


dlfivefifty:
>  The constructors were type piracy 

This is not true: it's not type piracy if the templated variables involve your own types (e.g., overloading `Eye` is fine because it involves `Ones`.)

Here the bigger issue is that `SubEye` is not closed under arithmetic: `2*eye(5,3)` will _not_ return a sparse matrix. It may be better as its own type.

Note another option is:
```julia
const Eye{T, EyeAxes, Axes<:NTuple{2, AbstractUnitRange}} = Union{Diagonal{T, Ones{T,1,Tuple{Axes}}}, SubArray{T,2,Diagonal{T, Ones{T,1,Tuple{Axes}}},Axes,false}}
```

iamed2:
> This is not true: it's not type piracy if the templated variables involve your own types (e.g., overloading Eye is fine because it involves Ones.)
I don't believe this is true; you wouldn't want to override `Array{MyType, 1}(::UndefInitializer, ::Int)` for instance, as this would prevent creation of arrays of that type.

> Here the bigger issue is that SubEye is not closed under arithmetic: 2*eye(5,3) will not return a sparse matrix. It may be better as its own type.
This is a very good point. I will make it its own type. I think `eye` as a function is a good idea as long as `Eye` and `SubEye` are different types.

dlfivefifty:
> you wouldn't want to override `Array{MyType, 1}(::UndefInitializer, ::Int)`

No you wouldn't. But it's still not "type piracy", which is about changing the behaviour of code not dependant on your package. Overloading `Array{MyType, 1}(::UndefInitializer, ::Int)` will not impact any other package, since it only applies to code that involves `MyType`.

> I think `eye` as a function is a good idea

The primary hesitation is that users might expect it to return a `Matrix`, like the old `eye`. 

I'd propose keeping `Eye` but having it be a union of the other types. Since these are immutable types, users can access the diagonal entries via `diag(A)`.

iamed2:
> The primary hesitation is that users might expect it to return a Matrix, like the old eye.

A number of linear algebra type changes have occurred as of 1.0, so I think people will be prepared for more. It's still an AbstractMatrix, after all. I do think that introducing `eye` on 0.7 would be a mistake (the deprecation still exists), which is why I only export it on 1.0. 

The downside of the union is that the `Axes` term would mean something different for the `Eye` part vs the `SubEye` part.

iamed2:
Can we continue this conversation? 

Should I put the SubEye type into its own PR?

dlfivefifty:
The issue is that the deprecation of `Eye` in favour of `eye`, and having `eye` do something different than it used to, seems wrong to me. 

It seems what we really want is a rectangular diagonal matrix:
```julia
struct RectDiagonal{T,V<:AbstractVector{T}, Axes} <: AbstractMatrix{T}
    diag::V
    axes::AX
end

const RectOrDiagonal{T, V, Axes} = Union{RectDiagonal{T, V, Axes}, Diagonal{T,V}}

const Eye{T, Axes} = RectOrDiagonal{T, Ones{T,1,Tuple{Axes}}}
```

iamed2:
> The issue is that the deprecation of Eye in favour of eye, and having eye do something different than it used to, seems wrong to me.

`eye` would be returning an `AbstractMatrix` equal to what it returned before, but with a different representation. I think that's a pretty standard switch.

> It seems what we really want is a rectangular diagonal matrix:

Can that live in FillArrays (at least for now) or should I attempt to get that into LinearAlgebra?

PR #49, opened 2018-12-03
Implement any/all methods

jlapeyre:
These are several methods for `any` and `all` and one method for
`isone`. Methods `any` and `all` are used in `Diagonal`, so they make
other methods efficient.

The tests included in this PR are designed to check the new
methods. But, there is no verification that the new methods are called
rather than the generic methods.

Because the methods for any/all/isone/iszero etc. are interrelated, a
benchmark suite would complement this commit.

### DIFF
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -3,7 +3,8 @@ module FillArrays
 using LinearAlgebra, SparseArrays
 import Base: size, getindex, setindex!, IndexStyle, checkbounds, convert,
                 +, -, *, /, \, sum, cumsum, maximum, minimum, sort, sort!,
-                axes, unique, allunique
+                any, all, axes, isone, iterate, unique, allunique
+
 import LinearAlgebra: rank, svdvals!
 
 import Base.Broadcast: broadcasted, DefaultArrayStyle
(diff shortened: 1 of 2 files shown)


src/FillArrays.jl
```diff
@@ -324,7 +324,39 @@ cumsum(x::AbstractFill{Bool}) = cumsum(convert(AbstractFill{Int}, x))
...
 unique(x::AbstractFill{T}) where T = isempty(x) ? T[] : T[getindex_value(x)]
 allunique(x::AbstractFill) = length(x) < 2
-           
+
+
+#########
+# any/all/isone/iszero
+#########
+
+function isone(AF::V) where {V <: AbstractFill{T,2}} where T
```
dlfivefifty: This signature is hard to read. This is preferable:
```julia
function isone(AF::AbstractFill{<:Any,2})
```

dlfivefifty:
Maybe doing a `d=0` test will fix the coverag?

src/FillArrays.jl
jlapeyre: Yes, that's better.

jlapeyre:
I'll look into the coverage.

jlapeyre:
It looks like the coverage is better now.

PR #50, opened 2018-12-05
Add methods that act as the identity on Eye

jlapeyre:
inv, permutedims, tril, triu, tril!, triu!

With this commit, these are all O(1) operations.


jlapeyre:
These PRs have been pretty fine grained. If they are too fine grained, I can group some that are closely related.  For example, other methods for `inv` might have been included here.

dlfivefifty:
Either way. Probably larger PRs would be less work for you.

PR #54, opened 2018-12-20
Rectangular Eye

iamed2:
I tried to follow the preferred design as much as possible. 

This code adds RectDiagonal as an implementation detail and as such does not export it. It could be developed further.

### DIFF
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -7,7 +7,7 @@ import Base: size, getindex, setindex!, IndexStyle, checkbounds, convert,
 
 import Base.\
 
-import LinearAlgebra: rank, svdvals!, tril, triu, tril!, triu!
+import LinearAlgebra: rank, svdvals!, tril, triu, tril!, triu!, diag
 
 import Base.Broadcast: broadcasted, DefaultArrayStyle, broadcast_shape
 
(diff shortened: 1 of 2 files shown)


dlfivefifty:
Looks good! Can we get the coverage up?

iamed2:
Looks like everything's covered now; there's one line that says it's not but I think that's an artifact of inlining as it's definitely covered.

src/FillArrays.jl
```diff
@@ -198,7 +198,59 @@ rank(F::Zeros) = 0
...
+diag(rd::RectDiagonal) = rd.diag
+
+for f in (:triu, :triu!, :tril, :tril!)
+    @eval ($f)(M::RectDiagonal) = M
+end
+
+
+const RectOrDiagonal{T,V,Axes} = Union{RectDiagonal{T,V,Axes}, Diagonal{T,V}}
+const SquareEye{T,Axes} = Diagonal{T,Ones{T,1,Tuple{Axes}}}
+const Eye{T,Axes} = RectOrDiagonal{T,Ones{T,1,Tuple{Axes}}}
```
ararslan: Looks like the change in type broke a test in InfiniteArrays (see [the JuliaCIBot results](https://juliarun-ci.s3.amazonaws.com/d0b0118e8c905b30b1d1ed15ef578d04ed55bbe5/InfiniteArrays_with_pull_request_of_FillArrays_on_julia_0_7.log)).

ararslan: I have a fix for it, I'll submit a PR to InfiniteArrays.

iamed2: Thanks!

PR #61, opened 2019-04-17
Tighter julia compat bounds

fredrikekre:


dlfivefifty:
This is what @StefanKarpinski's script https://github.com/JuliaLang/Pkg.jl/blob/master/bin/gen_project.jl produces.

PR #71, opened 2019-09-24
Add permutedims

dlfivefifty:
This adds `permutedims`.

I'm inclined to make this a minor tag: the chances that any user has mutated the result of `permutedims(::AbstractFill)` are probably zero, and if they had I feel like it's a "bug" that they assumed the result would be mutable. But one could also argue it's a breaking change.

 @sbromberger @willtebbutt @cstjean @fredrikekre any of you mind if its a minor tag?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FillArrays"
 uuid = "1a297f60-69ca-5386-bcde-b61e274b549b"
-version = "0.7"
+version = "0.7.1"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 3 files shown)


sbromberger:
No objection at all, especially since you're [not violating SemVer](https://semver.org/#spec-item-4). 😃 

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

dlfivefifty:
"Julia SemVar" is stricter...

sbromberger:
> "Julia SemVar" is stricter...

Interesting. I hadn't heard we have our own implementation of SemVer. Where is it documented?

dlfivefifty:
https://discourse.julialang.org/t/proposed-release-process-and-schedule/15623/31?u=dlfivefifty

sbromberger:
Hm, interesting. I'm still not worried about it in this case :)

PR #74, opened 2019-10-14
Add some specializations

kshyatt:
I got a migraine and didn't feel up to figuring out the broadcast semantics for the matrix versions, sorry :/. If anyone wants to edit this pr, go ahead. Should hopefully address https://github.com/FluxML/Zygote.jl/issues/261 .

### DIFF
--- a/src/fillalgebra.jl
+++ b/src/fillalgebra.jl
@@ -99,7 +99,27 @@ end
 *(a::Diagonal, b::Zeros{<:Any,1}) = mult_zeros(a, b)
 *(a::Diagonal, b::Zeros{<:Any,2}) = mult_zeros(a, b)
 
+*(a::Adjoint{T, <:StridedMatrix{T}},   b::Fill{T, 1}) where T = reshape(sum(conj.(parent(a)); dims=1) .* b.value, size(parent(a), 2))
+*(a::Transpose{T, <:StridedMatrix{T}}, b::Fill{T, 1}) where T = reshape(sum(parent(a); dims=1) .* b.value, size(parent(a), 2))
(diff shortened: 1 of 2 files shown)


willtebbutt:
@kshyatt happy to help out with the `Matrix * Matrix` versions.

Before that though, I notice the current implementation goes with `StridedMatrix` etc rather than `AbstractMatrix`, which I would have thought will still leave us with GPU-related issues. Might it be better to go with `AbstractMatrix` etc instead, or am I missing something?

kshyatt:
If you go with Abstract you get method ambiguity errors. I tried these on gpu and they seemed to work with scalar indexing off. CuArrays are subtypes of StridedArrays.

willtebbutt:
> CuArrays are subtypes of StridedArrays.

This doesn't seem to hold on Julia 1.3.0-rc3 and CuArrays v1.2.1. Am I missing something?

> If you go with Abstract you get method ambiguity errors

Good point -- Abstract probably isn't ideal then.

dlfivefifty:
We could overload both ` StridedMatrix ` and ` AbstractMatrix `

kshyatt:
@willtebbutt i'll check what versions of CuArrays and Julia I used, might have just been a mistake on my part :)

The method ambiguity comes because there are already specializations for `(A::AbstractFill{<:Any, 2}, Z::Zeros{<:Any, 1}` iiirc (quoting from memory here sorry if the signature is wrong).

kshyatt:
@willtebbutt I'm on the same Julia version and `CuArrays#master`, and I get:

```
julia> CuMatrix <: StridedArray
true

julia> a = cu(rand(10, 10));

julia> a isa CuMatrix
true

julia> a isa StridedArray
true

julia> a isa StridedMatrix
true
```

PR #78, opened 2019-11-01
Supporting Array Slicing

ho-oto:


dlfivefifty:
I pushed the `@inferred` tests but unfortunately they aren't passing. @ho-oto will you be able to fix the code so its type-stable?

dlfivefifty:
I believe this PR has been superseded by https://github.com/JuliaArrays/FillArrays.jl/pull/122

PR #87, opened 2020-02-20
implement conj, real, imag and zero

daanhb:
Downstream I noticed that the conjugate of a `Ones` with complex element type returned a more general `Fill` type. Chasing this and looking at the related definitions in base, I implemented `conj`, `real` and `imag` for Ones and Zeros.
As part of that, due to [this line](https://github.com/JuliaLang/julia/blob/3608c84e6093594fe86923339fc315231492484c/base/abstractarraymath.jl#L95) in Base, I implemented `zero(x)` as well for Ones, Zeros and Fill.

So, for example:
```julia
julia> imag(Ones(3))
3-element Zeros{Float64,1,Tuple{Base.OneTo{Int64}}}:
  ⋅
  ⋅
  ⋅
```
And the one that started it:
```julia
julia> conj(Zeros{ComplexF64}(3))
3-element Zeros{Complex{Float64},1,Tuple{Base.OneTo{Int64}}}:
      ⋅
      ⋅
      ⋅
```
And finally:
```julia
julia> zero(Fill(0.5, 3))
3-element Zeros{Float64,1,Tuple{Base.OneTo{Int64}}}:
  ⋅
  ⋅
  ⋅
```
There are added tests and a bumped version number.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FillArrays"
 uuid = "1a297f60-69ca-5386-bcde-b61e274b549b"
-version = "0.8.4"
+version = "0.8.5"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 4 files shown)


dlfivefifty:
Can you get the coverage up?

daanhb:
Not sure how - I checked and I'm hitting every line I've added, and the broadcasting ones twice (with different dimensions). The coverage report shows the changed lines with import statements, and some lines where my editor (Atom) apparently automatically removed trailing whitespace.

daanhb:
It could be that the added lines changed the code paths in a way that affects other tests. Let me check that.

daanhb:
Looks better now?

PR #89, opened 2020-03-28
Fix #85

### LINKED ISSUE #85: Support Colon in reshape overloads
https://github.com/JuliaArrays/FillArrays.jl/blob/60235cbb8ad2bf1cf3b781cb9f29902f6f793881/src/FillArrays.jl#L168-L169 these don't support `Colon`, so something like `reshape(A,:,5)` will cause this to create a `ReshapedArray`.

dlfivefifty:
I also now only test on the stable release (1.0) and current release (1.4) to minimise CI burden. In any case I think it's reasonable to only support these two releases (if it were to break on 1.2 we would tell the user to upgrade to 1.4)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,6 @@ os:
   - osx
 julia:
   - 1.0
-  - 1.1
-  - 1.2
-  - 1.3
   - 1.4
   - nightly
 matrix:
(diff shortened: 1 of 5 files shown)


dlfivefifty:
The test failure is due to appveyor issues.

PR #92, opened 2020-04-29
Overload lmul!/rmul!

dlfivefifty:


src/fillalgebra.jl
```diff
@@ -211,3 +211,20 @@ norm2(a::AbstractFill) = sqrt(length(a))*norm(getindex_value(a))
...
 normMinusInf(a::AbstractFill) = norm(getindex_value(a))
+
+
+###
+# lmul!/rmul!
+###
+
+function lmul!(x::Number, z::AbstractFill)
+    λ = getindex_value(z)
+    x*λ == λ || throw(ArgumentError("Cannot scale by $x"))
```
willtebbutt, line 222: Could you add a comment at the top of this method explaining the rationale for this check? Definitely wouldn't be obvious to me why this is reasonable if I were perusing the code.

PR #103, opened 2020-07-02
Add constructors for Fill(::Fill) and varients

dlfivefifty:
Some parts of Base/StdLib have calls like `typeof(A)(B)` for converting a matrix `B` to that of type `A`. 

This is a bit subtle for `Fill`: `Fill(x)` creates a 0-dimensional array whose only entry is `x`, so should `Fill(x::Fill)` return `x` or a `Fill{<:Fill,0}`? I've left it as the latter for now by only overriding `Fill{T}(F::Fill{T})`, but perhaps this needs more thought.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FillArrays"
 uuid = "1a297f60-69ca-5386-bcde-b61e274b549b"
-version = "0.8.11"
+version = "0.8.12"
 
 [deps]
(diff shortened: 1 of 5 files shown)


dlfivefifty:
This also "corrects" behaviour of `Zeros(::AbstractArray{T})` to now return a `Zeros{T}`, not a `Zeros{Float64}`.

Issue #95, opened 2020-05-02
no docstrings for Zeros, Ones, Fill


CarloLucibello:
e.g.
```julia
help?> Ones
search: Ones ones leading_ones trailing_ones countlines count_ones to_indices RoundNearest RoundNearestTiesUp RoundNearestTiesAway count_zeros propertynames one oneunit isone exponent

  No documentation found.

  FillArrays.Ones is of type UnionAll.

  Summary
  ≡≡≡≡≡≡≡≡≡

  struct UnionAll <: Type{T}

  Fields
  ≡≡≡≡≡≡≡≡

  var  :: TypeVar
  body :: Any

  Supertype Hierarchy
  ≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡

  UnionAll <: Type{T} <: Any
```

dlfivefifty:
👍 A PR for this would be great.

JeffFessler:
Probably this can be closed now due to #109

PR #109, opened 2020-08-01
Concise version of "show"

JeffFessler:
The default way that `AbstractFill` types are displayed at the REPL produces reams of repeated output values for large dimensions, despite a `Fill` containing essentially just one value.
Simply displaying the dimensions, the axes and the value suffices to describe concisely a `Fill` object.
This PR simply provides a concise `show` method for any `Fill` type.

To see the benefit, compare the REPL output of
`Ones(100,100,100)`
for this version, which becomes
`100×100×100 Ones{Float64,3,Tuple{Base.OneTo{Int64},Base.OneTo{Int64},Base.OneTo{Int64}}} = 1.0`
versus the original version (that is too long to paste here).

I also added a few docstrings to partially address #95.

### DIFF
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -4,7 +4,8 @@ using LinearAlgebra, SparseArrays
 import Base: size, getindex, setindex!, IndexStyle, checkbounds, convert,
     +, -, *, /, \, diff, sum, cumsum, maximum, minimum, sort, sort!,
     any, all, axes, isone, iterate, unique, allunique, permutedims, inv,
-    copy, vec, setindex!, count, ==, reshape, _throw_dmrs, map, zero
+    copy, vec, setindex!, count, ==, reshape, _throw_dmrs, map, zero,
+    show
 
 import LinearAlgebra: rank, svdvals!, tril, triu, tril!, triu!, diag, transpose, adjoint, fill!,
     norm2, norm1, normInf, normMinusInf, normp, lmul!, rmul!, diagzero, AbstractTriangular
@@ -551,5 +552,18 @@ Base.print_matrix_row(io::IO,
                  }, A::Vector,
         i::Integer, cols::AbstractVector, sep::AbstractString) =
         axes_print_matrix_row(axes(X), io, X, A, i, cols, sep)
+
+
+"""
+    show(io::IO, x::AbstractFill)
+    show(io::IO, ::MIME"text/plain", x::AbstractFill)
+Display concise description.
+"""
+Base.show(io::IO, x::AbstractFill) =
+    print(io, "$(summary(x)) = $(getindex_value(x))")
+
+function Base.show(io::IO, ::MIME"text/plain", x::AbstractFill)
+    show(io, x)
+end
 
 end # module
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -1,3 +1,4 @@
+""" `FillArrays` module to lazily represent matrices with a single value """
 module FillArrays
 
 using LinearAlgebra, SparseArrays
@@ -16,6 +17,7 @@ import Base.Broadcast: broadcasted, DefaultArrayStyle, broadcast_shape
 
 export Zeros, Ones, Fill, Eye
 
+""" `AbstractFill{T, N, Axes} <: AbstractArray{T, N}` """
 abstract type AbstractFill{T, N, Axes} <: AbstractArray{T, N} end
 
 ==(a::AbstractFill, b::AbstractFill) = axes(a) == axes(b) && getindex_value(a) == getindex_value(b)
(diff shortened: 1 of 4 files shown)


JeffFessler:
Apparently this change causes the function `Base.replace_in_print_matrix` to go uncovered in the tests.
Probably that could be restored by adding a test for some non-`Fill` type like `Eye(2)`, if covering it is important to package maintainers. (Personally I would prefer a more concise display of `Eye` too, but that might be more controversial so I didn't include that in this PR.)

test/runtests.jl
```diff
@@ -91,6 +91,8 @@ import FillArrays: AbstractFill, RectDiagonal, SquareEye
                 Fill{Float32}(one(Float32),5,5)
 
         @test Fill{T}(F) ≡ Fill{T,2}(F) ≡ typeof(F)(F) ≡ F
+
+        show(devnull, MIME("text/plain"), F)
```
dlfivefifty, line 95: What is this call for?

src/FillArrays.jl
```diff
@@ -552,4 +560,17 @@ Base.print_matrix_row(io::IO,
         i::Integer, cols::AbstractVector, sep::AbstractString) =
         axes_print_matrix_row(axes(X), io, X, A, i, cols, sep)
 
+
+"""
+    show(io::IO, x::AbstractFill)
+    show(io::IO, ::MIME"text/plain", x::AbstractFill)
+Display concise description.
+"""
```
dlfivefifty, line 568: This doc-string is not needed and will just add noise to `? show`

src/FillArrays.jl
```diff
@@ -50,6 +53,7 @@ rank(F::AbstractFill) = iszero(getindex_value(F)) ? 0 : 1
 IndexStyle(::Type{<:AbstractFill{<:Any,N,<:NTuple{N,Base.OneTo{Int}}}}) where N = IndexLinear()
 
 
+""" `Fill{T, N, Axes} <: AbstractFill{T, N, Axes}` (lazy `fill` with axes)"""
```
dlfivefifty, line 56: This is simealtanteously too terse and too verbose: the `<: AbstractFill{T, N, Axes}` is unneeded because a user can call `supertype` if they want that information. Try to make it look like the docs for lazy types in base, e.g., 
```julia
help?> UnitRange
search: UnitRange AbstractUnitRange UnitUpperTriangular UnitLowerTriangular

  UnitRange{T<:Real}

  A range parameterized by a start and stop of type T, filled with elements
  spaced by 1 from start until stop is exceeded. The syntax a:b with a and b
  both Integers creates a UnitRange.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> collect(UnitRange(2.3, 5.2))
  3-element Array{Float64,1}:
   2.3
   3.3
   4.3
  
  julia> typeof(1:10)
  UnitRange{Int64}
```

src/FillArrays.jl
```diff
@@ -1,10 +1,12 @@
+""" `FillArrays` module to lazily represent matrices with a single value """
```
dlfivefifty, line 1: Delete this: I've never seen a package have a doc string before so this is not standard.

src/FillArrays.jl
```diff
@@ -15,6 +17,7 @@ import Base.Broadcast: broadcasted, DefaultArrayStyle, broadcast_shape
 
 export Zeros, Ones, Fill, Eye
 
+""" `AbstractFill{T, N, Axes} <: AbstractArray{T, N}` """
```
dlfivefifty, line 20: This docstring is not useful so either delete it, or change it to something like
```
AbstractFill{T, N, Axes} <: AbstractArray{T, N}

Supertype for lazy array types whose entries are all equal to constant.
```

test/runtests.jl
JeffFessler: This call is to provide code coverage for the new `show` method.
Without some call like this, the code coverage for the package will be reduced.

src/FillArrays.jl
JeffFessler: Ok, I will delete it.
Just for the record this is standard enough that it is documented in the manual:
https://docs.julialang.org/en/v1/manual/documentation/index.html#Modules-1
but this was my first time trying it out - just trying to be helpful...

src/FillArrays.jl
JeffFessler: Oops, that was pretty useless.

src/FillArrays.jl
JeffFessler: Done.

src/FillArrays.jl
JeffFessler: Cut.

JeffFessler:
Thanks for the feedback.  I think I addressed the requests.

src/FillArrays.jl
dlfivefifty: Ok in that case leave it (never seen it before...)

JeffFessler: Ok, I put it back.
(I haven't seen it much either and a good README for a package seems much more important.)

JeffFessler:
I'd recommend squashing the commits on merge (probably goes without saying)...

dlfivefifty:
Can you bump the patch number so I can tag a new version?

PR #110, opened 2020-08-03
Add `Trues` type and methods

JeffFessler:
The "show business" (pun intended) at #109 was just a warm-up.  This PR is the one that matters.
Here's a bit of background.  Unaware of `FillArrays.jl`, I created a somewhat similar package [ConstantArrays.jl](https://github.com/JeffFessler/ConstantArrays.jl), and only later learned about the duplication.
One of my main motivations for writing `ConstantArrays` was to have very efficient versions of `getindex` and `setindex!` for "masks" that are `trues(dim)`.
I have timing tests that show the methods for `getindex` and `setindex!` in this PR reduce the time from 1.6ms to 28ns for `getindex` and from 2.3ms to 0.6ms for `setindex!` using `Trues`:
https://github.com/JeffFessler/ConstantArrays.jl/blob/master/test/time.jl
If this PR goes through then I can delete the otherwise redundant `ConstantArrays` and use `FillArrays` hereafter.
BTW, I don't have a use case for `Falses` right now, but I included it for symmetry with `Trues`.
There is one possible test for matching `axes` that I included but left commented out because I don't think it is relevant when indexing using `Bool` arrays, but I am not that familiar with `axes` so I left it there in case one of you experts says it is needed.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FillArrays"
 uuid = "1a297f60-69ca-5386-bcde-b61e274b549b"
-version = "0.8.13"
+version = "0.8.14"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FillArrays"
 uuid = "1a297f60-69ca-5386-bcde-b61e274b549b"
-version = "0.8.14"
+version = "0.9"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -127,11 +127,11 @@ function getindex(F::Fill, kj::Vararg{AbstractVector{II},N}) where {II<:Integer,
 end
 
 function getindex(A::Fill, kr::AbstractVector{Bool})
-   length(A) == length(kr) || throw(DimensionMismatch())
+   length(A) == length(kr) || throw(DimensionMismatch("lengths must match"))
    Fill(getindex_value(A), count(kr))
 end
 function getindex(A::Fill, kr::AbstractArray{Bool})
-   size(A) == size(kr) || throw(DimensionMismatch())
+   size(A) == size(kr) || throw(DimensionMismatch("sizes must match"))
    Fill(getindex_value(A), count(kr))
 end
 
@@ -235,11 +235,11 @@ for (Typ, funcs, func) in ((:Zeros, :zeros, :zero), (:Ones, :ones, :one))
             $Typ{T}(length.(kj))
         end
         function getindex(A::$Typ{T}, kr::AbstractVector{Bool}) where T
-            length(A) == length(kr) || throw(DimensionMismatch())
+            length(A) == length(kr) || throw(DimensionMismatch("lengths must match"))
             $Typ{T}(count(kr))
         end
         function getindex(A::$Typ{T}, kr::AbstractArray{Bool}) where T
-            size(A) == size(kr) || throw(DimensionMismatch())
+            size(A) == size(kr) || throw(DimensionMismatch("sizes must match"))
             $Typ{T}(count(kr))
         end
 
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -529,6 +529,7 @@ count(f, x::AbstractFill) = f(getindex_value(x)) ? length(x) : 0
 
 include("fillalgebra.jl")
 include("fillbroadcast.jl")
+include("trues.jl")
 
 ##
 # print
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -13,7 +13,7 @@ import Base.Broadcast: broadcasted, DefaultArrayStyle, broadcast_shape
 
 
 
-export Zeros, Ones, Fill, Eye
+export Zeros, Ones, Fill, Eye, Trues, Falses
 
 abstract type AbstractFill{T, N, Axes} <: AbstractArray{T, N} end
 
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -159,11 +159,11 @@ function getindex(F::Fill, kj::Vararg{AbstractVector{II},N}) where {II<:Integer,
 end
 
 function getindex(A::Fill, kr::AbstractVector{Bool})
-   length(A) == length(kr) || throw(DimensionMismatch("lengths must match"))
+   length(A) == length(kr) || throw(DimensionMismatch())
    Fill(getindex_value(A), count(kr))
 end
 function getindex(A::Fill, kr::AbstractArray{Bool})
-   size(A) == size(kr) || throw(DimensionMismatch("sizes must match"))
+   size(A) == size(kr) || throw(DimensionMismatch())
    Fill(getindex_value(A), count(kr))
 end
 
--- a/src/trues.jl
+++ b/src/trues.jl
@@ -22,16 +22,23 @@ const Falses = Zeros{Bool, N, Axes} where {N, Axes}
 
 
 # y[mask] = x when mask isa Trues (cf y[:] = x)
-function Base.setindex!(y::AbstractArray{T,N}, x, mask::Trues{N}) where {T,N}
-    @boundscheck axes(y) == axes(mask) || throw(BoundsError(y, mask))
-    @boundscheck axes(x) == axes(mask) || throw(DimensionMismatch(
-        "tried to assign $(length(x)) elements to $(length(y)) destinations"))
-    @boundscheck checkbounds(y, mask)
-    copyto!(y, x)
+# Supported here only for arrays with standard OneTo axes.
+function Base.setindex!(y::AbstractArray{T,N}, x,
+    mask::Trues{N, NTuple{N,Base.OneTo{Int}}},
+) where {T,N}
+    if axes(x) isa NTuple{N,Base.OneTo{Int}} &&
+       axes(y) isa NTuple{N,Base.OneTo{Int}}
+        @boundscheck size(y) == size(mask) || throw(BoundsError(y, mask))
+        @boundscheck size(x) == size(mask) || throw(DimensionMismatch(
+            "tried to assign $(length(x)) elements to $(length(y)) destinations"))
+        @boundscheck checkbounds(y, mask)
+        return copyto!(y, x)
+    end
+    return setindex!(y, x, trues(size(mask))) # fall back on usual setindex!
 end
 
 # x[mask] when mask isa Trues (cf x[trues(size(x))] or x[:])
-# Supported here only for a mask with standard OneTo axes.
+# Supported here only for arrays with standard OneTo axes.
 function Base.getindex(x::AbstractArray{T,N},
     mask::Trues{N, NTuple{N,Base.OneTo{Int}}},
 ) where {T,N}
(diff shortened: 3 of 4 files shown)


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "FillArrays"
 uuid = "1a297f60-69ca-5386-bcde-b61e274b549b"
-version = "0.8.13"
+version = "0.8.14"
```
dlfivefifty, line 3: Should be 0.9 since you export new types

src/trues.jl
```diff
@@ -0,0 +1,37 @@
...
+const Trues = Ones{Bool, N, Axes} where {N, Axes}
+
+
+""" `Falses = Zeros{Bool, N, Axes}` (see `Trues`) """
+const Falses = Zeros{Bool, N, Axes} where {N, Axes}
+
+
+# y[mask] = x when mask isa Trues (cf y[:] = x)
+function Base.setindex!(y::AbstractArray{T,N}, x, mask::Trues{N}) where {T,N}
+    @boundscheck size(x) == size(mask) == size(y) || throw(DimensionMismatch())
```
dlfivefifty, line 26: Need to put a string description in `DimensionMismatch`. Also, perhaps `axes(x) == axes(mask) == axes(y)` is more general?

src/trues.jl
```diff
@@ -0,0 +1,37 @@
...
+    @boundscheck checkbounds(x, mask)
+#   @boundscheck axes(mask) == axes(x) || throw("axes mismatch")
+    copyto!(y, x)
+end
+
+# x[mask] when mask isa Trues (cf x[trues(size(x))] or x[:])
+function Base.getindex(x::AbstractArray{T,D}, mask::Trues{D}) where {T,D}
+    @boundscheck size(x) == size(mask) || throw(DimensionMismatch())
```
dlfivefifty, line 34: Here too.

src/trues.jl
```diff
@@ -0,0 +1,37 @@
...
+
+""" `Falses = Zeros{Bool, N, Axes}` (see `Trues`) """
+const Falses = Zeros{Bool, N, Axes} where {N, Axes}
+
+
+# y[mask] = x when mask isa Trues (cf y[:] = x)
+function Base.setindex!(y::AbstractArray{T,N}, x, mask::Trues{N}) where {T,N}
+    @boundscheck size(x) == size(mask) == size(y) || throw(DimensionMismatch())
+    @boundscheck checkbounds(x, mask)
+#   @boundscheck axes(mask) == axes(x) || throw("axes mismatch")
```
dlfivefifty, line 28: Delete this comment.

Project.toml
JeffFessler: ✓

src/trues.jl
JeffFessler: ✓
(found a few others too...)

src/trues.jl
JeffFessler: ✓

src/trues.jl
```diff
@@ -0,0 +1,38 @@
...
+    @boundscheck axes(x) == axes(mask) == axes(y) ||
+        throw(DimensionMismatch("axes must match for setindex!"))
+    @boundscheck checkbounds(y, mask)
+    copyto!(y, x)
+end
+
+# x[mask] when mask isa Trues (cf x[trues(size(x))] or x[:])
+function Base.getindex(x::AbstractArray{T,D}, mask::Trues{D}) where {T,D}
+    @boundscheck axes(x) == axes(mask) ||
+        throw(DimensionMismatch("axes must match for getindex"))
```
dlfivefifty, line 35: This should be a `BoundsError` to match `Array`:
```julia
julia> a = randn(5)
5-element Array{Float64,1}:
  0.5894229873306978
  1.1656206737590917
  0.314932122096189
 -0.8914119854410714
  0.7630084129166993

julia> a[[true,true]]
ERROR: BoundsError: attempt to access 5-element Array{Float64,1} at index [Bool[1, 1]]
Stacktrace:
 [1] throw_boundserror(::Array{Float64,1}, ::Tuple{Base.LogicalIndex{Int64,Array{Bool,1}}}) at ./abstractarray.jl:541
 [2] checkbounds at ./abstractarray.jl:506 [inlined]
 [3] _getindex at ./multidimensional.jl:742 [inlined]
 [4] getindex(::Array{Float64,1}, ::Array{Bool,1}) at ./abstractarray.jl:1060
 [5] top-level scope at REPL[2]:1
```

src/trues.jl
```diff
@@ -0,0 +1,38 @@
...
+
+
+""" `Falses = Zeros{Bool, N, Axes}` (see `Trues`) """
+const Falses = Zeros{Bool, N, Axes} where {N, Axes}
+
+
+# y[mask] = x when mask isa Trues (cf y[:] = x)
+function Base.setindex!(y::AbstractArray{T,N}, x, mask::Trues{N}) where {T,N}
+    @boundscheck axes(x) == axes(mask) == axes(y) ||
+        throw(DimensionMismatch("axes must match for setindex!"))
```
dlfivefifty, line 27: This should be a `BoundsError` to match `Array`:
```julia
julia> setindex!(a, [1,2], [true,true])
ERROR: BoundsError: attempt to access 5-element Array{Float64,1} at index [Bool[1, 1]]
Stacktrace:
 [1] throw_boundserror(::Array{Float64,1}, ::Tuple{Base.LogicalIndex{Int64,Array{Bool,1}}}) at ./abstractarray.jl:541
 [2] checkbounds at ./abstractarray.jl:506 [inlined]
 [3] _setindex! at ./multidimensional.jl:784 [inlined]
 [4] setindex!(::Array{Float64,1}, ::Array{Int64,1}, ::Array{Bool,1}) at ./abstractarray.jl:1153
 [5] top-level scope at REPL[3]:1
```

test/runtests.jl
```diff
@@ -1012,4 +1012,16 @@ end
...
+@testset "Trues" begin
+    @test Trues(2,3) == Trues((2,3)) == trues(2,3)
+    @test Falses(2,3) == Falses((2,3)) == falses(2,3)
+    dim = (4,5)
+    mask = Trues(dim)
+    x = randn(dim)
+    @test x[mask] == vec(x) # getindex
+    y = similar(x)
+    y[mask] = x # setindex!
+    @test y == x
```
dlfivefifty, line 1026: Add `@test_throws BoundsError ...` to test case where axes don't match.

src/trues.jl
JeffFessler: done

test/runtests.jl
JeffFessler: done, only took 4 commits to get it right (i hope)...

dlfivefifty:
The error messages are still wrong: they need to match `Array` and call `checkbounds` to get the message correct. 

But having thought about this a bit more, I'm not convinced that  overloading `getindex` is the right way to do this. I think we may want to overload `to_indices`, see 

https://github.com/JuliaLang/julia/blob/4c1e3c039775e56c72a4a638db0a8a20f4972dd3/base/multidimensional.jl#L719

That is, perhaps something like this, which just returns the corresponding axis:
```julia
Base.to_indices(A::AbstractArray{N}, inds, I::Tuple{Trues{N}}) where N = Base.to_indices(A, inds, (:,))
```
The benefit is you don't need to worry about bounds checking. The question though is whether there are negative consequences of this.

JeffFessler:
I need help understanding / fixing the problem with the error messages.
Here are the errors messages for `Trues` and the corresponding messages for `trues` -- they look essentially identical to me:
```
ones(3)[trues(2)]
ERROR: BoundsError: attempt to access 3-element Array{Float64,1} at index [Bool[1, 1]]

ones(3)[Trues(2)]
ERROR: BoundsError: attempt to access 3-element Array{Float64,1} at index [true, true]


setindex!(ones(3), zeros(3), trues(2))
ERROR: BoundsError: attempt to access 3-element Array{Float64,1} at index [Bool[1, 1]]

setindex!(ones(3), zeros(3), Trues(2))
ERROR: BoundsError: attempt to access 3-element Array{Float64,1} at index [true, true]


setindex!(ones(2), zeros(3), trues(2))
ERROR: DimensionMismatch("tried to assign 3 elements to 2 destinations")

setindex!(ones(2), zeros(3), Trues(2))
ERROR: DimensionMismatch("tried to assign 3 elements to 2 destinations")
```

Sorry, but I do not see how to use `to_indices` here because the example you show uses a `Tuple` of 1D `Trues` but the method I need uses a single `Trues` argument, like `ones(3)[Trues(3)]`, not a Tuple.  Nevertheless, I tried the suggestion, but my timing tests showed that it did not provide any speedup over how `getindex` is handled by the existing code in `FillArrays`.  If interested you can see the details in this test/timing file:
https://github.com/JeffFessler/ConstantArrays.jl/blob/master/test/to_indices.jl

Perhaps your concern has to do with types that use non-`OneTo` axes.  I need this extension only for arrays with `OneTo` axes.  The only non-`OneTo` axes I recall ever using are those in `FFTViews` and I did a little test with one of those types and found that it did not support using an array of Bools as indices.  (The little test is in the comments at the bottom of the link above.)

Thus, I've pushed a new commit where I've modified the `getindex` method to work *only* with `OneTo` types, and revert to whatever existing `getindex` method there is (if any) for other types.  This simplifies the bounds checking to just a size check.  If this restriction to `OneTo` types looks acceptable to you then I will modify `setindex!` with the same strategy.

I don't understand enough about axes to try to generalize this to non-`OneTo` types anyway, especially when it comes to using Bool indices for such types.

src/FillArrays.jl
```diff
@@ -159,11 +159,11 @@ function getindex(F::Fill, kj::Vararg{AbstractVector{II},N}) where {II<:Integer,
 end
 
 function getindex(A::Fill, kr::AbstractVector{Bool})
-   length(A) == length(kr) || throw(DimensionMismatch())
+   length(A) == length(kr) || throw(DimensionMismatch("lengths must match"))
```
dlfivefifty, line 162: This is the message that looks wrong to me. E.g.
```julia
Fill(5,10)[trues(9)]
```

dlfivefifty:
OK we can forget `to_indices` for now, just thought it might be cleaner. I've added a comment of the error message that looks wrong

src/FillArrays.jl
JeffFessler: Oh, this was a foolish attempt on my part to try to be helpful by sticking in a String in another location after you pointed out that I was missing a String in the code I had added.  I have reverted this back to the original string-less version because it will be more efficient for you to update it the way you want instead of me trying to guess.  I should not have mixed unrelated changes to other files in this PR...

JeffFessler:
Ok, I have updated `setindex!` so that it is also applicable solely to `OneTo` cases, which seems cleaner, and I've timed it and it still provides the performance benefit expected.

dlfivefifty:
You still need to fix the errors for `getindex(A::$Typ{T}, kr::AbstractVector{Bool}) where T`

JeffFessler:
I had replied to that request right below where you suggested it, and then I clicked "resolve conversation" - which in retrospect probably was another freshman mistake on my part.  It's unclear to me who is supposed to click it.  Now I've "unresolved it" so perhaps my previous reply will be more visible.

The short version of my reply is "not my job because that was not code that I added and I shouldn't have messed with it in the first place in this PR."  So I reverted it back to the original version that had no string. 

Now the weird thing is that github is not displaying correctly the fact that I reverted it in the "Files changed" tabl.
If you go directly to the file you will see that the non-compliant string is gone:
https://github.com/JeffFessler/FillArrays.jl/blob/63968e46b2098832497924874e1aef62dee16d19/src/FillArrays.jl#L162
But (at least for me) under "Files changed" the difference does not reflect the fact that I reverted it.

dlfivefifty:
OK fair enough (though next time it might be worth considering just to fix it as a "service")

JeffFessler:
Thanks for the merge.  I am happy to do a "service" when I am confident about the code, like I did with the docstrings, but I am not confident about @inbounds and @boundscheck details.

PR #111, opened 2020-08-07
Add new types to readme

JeffFessler:
I forgot to update the README with the new types.  Fixed a couple typos while in there, as a service :)

An unintended consequence of the new `show` is that the examples in the README don't quite match the actual code output because they show the older more verbose output.  I used the new output format for the `Trues` and `Falses` additions, but I left the other ones untouched.  if you want me to update the README to match the actual output of show in the current code I can.  Personally I think it may be more intuitive in the README the way it is, but your call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,9 @@
 [![Build Status](https://travis-ci.org/JuliaArrays/FillArrays.jl.svg?branch=master)](https://travis-ci.org/JuliaArrays/FillArrays.jl)
 [![codecov](https://codecov.io/gh/JuliaArrays/FillArrays.jl/branch/master/graph/badge.svg)](https://codecov.io/gh/JuliaArrays/FillArrays.jl)
 
-Julia package to lazily representing matrices filled with a single entry,
-as well as identity matrices.  This package exports the following types: `Eye`,
-`Fill`, `Ones`, and `Zeros`.
+Julia package to lazily represent matrices filled with a single entry,
+as well as identity matrices.  This package exports the following types:
+`Eye`, `Fill`, `Ones`, `Zeros`, `Trues` and `Falses`.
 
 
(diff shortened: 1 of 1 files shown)


dlfivefifty:
It should be consistent with the actual output. To be honest I think the output could be refined: It's not immediately clear that
```julia
julia> Fill(1,10)
10-element Fill{Int64,1,Tuple{Base.OneTo{Int64}}} = 1
```
means the entries are all `1`. And the template information is also too much when it's the default. It might actually be preferable just to return `Fill(1,10)`, or possibly `10-element Fill with entries all 1`, which for matrices would be `10×5 Fill with entries all 1`.

JeffFessler:
I'd be happy to modify `show` (and update readme to match) if you specify exactly which output format you want.  I agree that `= 1` was a lazy shorthand. The "with entries all 1" suggestion looks good to me.  Most types I've seen use `summary()` to show the type information (even if perhaps a bit too much), and it was very easy to code using `summary()`,  but anything will be better than the original screenfulls of values so I'll gladly make it whichever way you want.

dlfivefifty:
Maybe add an example like
```julia
julia> Array(Fill(2,2,3))
2×3 Array{Int64,2}:
 2  2  2
 2  2  2
```
to show the full array?

JeffFessler:
Done.  I don't know if there is a standard in Julia for showing the axes, but I feel that
`... indices 0:2×-1:0 ...`
would be easier to read as
`... indices (0:2)×(-1:0) ...`
or at least
`... indices 0:2 × -1:0 ...`
(Maybe this comment belongs with #113)

JeffFessler:
@dlfivefifty I think the updated readme is ready (if you are fine with the way the indices look).

PR #113, opened 2020-08-10
size -> axes in broadcast_shape

dlfivefifty:
This calls `axes` instead of `size` in broadcasting to work better with BlockArrays.jl.

I've also improved the printing to be more descriptive and less verbose, @JeffFessler are you happy with this change?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FillArrays"
 uuid = "1a297f60-69ca-5386-bcde-b61e274b549b"
-version = "0.9.1"
+version = "0.9.2"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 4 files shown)


JeffFessler:
It would be nice to show the `eltype` for `Fill` like it does for `Zeros` (and for typical Julia Arrays) both because it is useful and to make the following outputs more similar:

```julia
julia> Zeros(3,4)
3×4 Zeros{Float64}

julia> Fill(Int32(0.), (3,4))
3×4 Fill: entries equal to 0

julia> Fill(Float32(0.), (3,4))
3×4 Fill: entries equal to 0.0
```

Perhaps:
`3×4 Fill{Float32}: entries equal to 0.0`

I'm not sure why it is not showing `0.0f0` here but that's secondary to showing the `eltype`.

dlfivefifty:
I didn't bother showing the `eltype` as I thought it was apparent from the entry, but your example proves me wrong!

JeffFessler:
Looks good!  I tried to update #111 to match.  Not completely sure about it...

PR #115, opened 2020-08-16
Fix broadcasting when FIll is array-valued

dlfivefifty:


dlfivefifty:
For some reason codecov is not updating, so the coverage decrease is misleading.

PR #122, opened 2020-10-21
Add special broadcast routines for +

dlfivefifty:


dlfivefifty:
This is a breaking change as conceivably one might have done
```julia
a = Zeros(5,5)[:,1]
a[1] = 2
```
downstream, which no longer works because now `a isa Zeros`.

PR #131, opened 2020-12-17
Avoid allocation in FillArrays.Zero addition

crstnbr:
`X` is a dense matrix, `Z` is a `FillArrays.Zero`.

Before this PR:

```Julia
julia> @btime $X+$Z
  70.028 ns (1 allocation: 896 bytes)
```

Afterwards:
```julia
julia> @btime $X+$Z
  4.327 ns (0 allocations: 0 bytes)
```

(Here `X` is a `Matrix{Float64}` and `Z` is a `FillArrays.Zero{Float64}` but the PR addresses all cases where the type parameters are equal.)


dlfivefifty:
` + `   should allocate as a user may modify the returned array. Look at for example ` A^1 ` which calls `copy `

crstnbr:
Fair enough.

Issue #104, opened 2020-07-06
Unify `Ones` and `Zeros` as `SFill{0}` and `SFill{k}`?


dlfivefifty:
It might make the code cleaner to do:
```julia
struct SFill{k, T, N, Axes} <: AbstractFill{T,N}
   axes::Axes
end

getindex_value(F::SFill{k,T}) where {k,T} = convert(T,k)

const Ones{T,N,Axes} = SFill{true,T,N,Axes}
const Zeros{T,N,Axes} = SFill{false,T,N,Axes}
```
It would then be possible to support other special cases. Though whether this machinery is worth it is not clear.

c42f:
Coming here from https://github.com/JuliaArrays/StaticArrays.jl/issues/815.

I'll let others debate about whether the machinery is worth it :) A small note - I'm not sure you want the value `k` in the type if it can be a non bits type? Though it seems fine for the `Ones` and `Zeros` you've defined here.

dlfivefifty:
We already have `Fill` when the value is not in the type...

A more pressing case is where the length is static. But it seems this might be better done via `Axes`? If only there were a `SOneTo`.

oschulz:
Coming from #176 (compatibility with Static.jl), might

```julia
const Ones{T,N,Axes} = Fill{Static.True,N,Axes}
const Zeros{T,N,Axes} = Fill{Static.False,N,Axes}
```

not be an elegant alternative?

dlfivefifty:
Where did `T` go?

oschulz:
#140 could then be extended to enable `Fill`-to-`Fill` vcat for arbitrary static values, not just one and zero, we could have

```
vcat(Fill(static(4.2), 5), Fill(static(4.2), 5)) === Fill(static(4.2), 10)
```

oschulz:
> Where did T go?

Ooops, indeed! :-) That's a problem, because there's no common static supertype across all value types in Static.jl. But we could do

```julia
struct SFill{k, T, N, Axes} <: AbstractFill{T,N}
   axes::Axes
end

getindex_value(F::SFill{k,T}) where {k,T} = convert(T,static(k))

const Ones{T,N,Axes} = SFill{true,T,N,Axes}
const Zeros{T,N,Axes} = SFill{false,T,N,Axes}
```

after SciML/Static.jl#50 is fixed.

dlfivefifty:
Except then you lose the propagation of static values so there's no point...

[...]

PR #140, opened 2021-02-21
Add cat, vcat and hcat

### LINKED ISSUE #23: Add hcat, vcat, and cat methods
I recently had a need for these functions. I wrote a quick method (below) for my particular case, but I would be better if they were officially supported.

```julia
function Base.hcat(x::Zeros{T,2}...) where T <: Real
    all(xx -> xx.size[1] == x[1].size[1], x) || throw(ArgumentError("mismatch in dimension 1"))
    n = sum(xx.size[2] for xx in x)

    return Zeros(x[1].size[1],n)
end
```

DrChainsaw:
Fixes #23 in case the offer from 2018 still stands :)

There where a few more edge cases than I had first thought and I hope the testing covers them all. 

There are a couple of cases where it is not possible to return a `Fill` or `Ones` and then a normal `Array` is returned. Same goes for mixed types (e.g. `vcat(Fill(0, 3), Zeros(4))` as I don't think there exist a feasible way to handle those cases. 

Just close with prejudice if this is not wanted any more.

I guess this is kind of breaking as there might be code out there which (perhaps inadvertently) relies on concatenation returning an `Array`.

### DIFF
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -589,6 +589,7 @@ end
 include("fillalgebra.jl")
 include("fillbroadcast.jl")
 include("trues.jl")
+include("fillcat.jl")
 
 ##
 # print
--- a/src/fillcat.jl
+++ b/src/fillcat.jl
@@ -5,7 +5,7 @@ function Base.cat_t(::Type{T}, fs::Fill...; dims) where T
 
     catdims = Base.dims2cat(dims)
 
-    # Note, when dims is a tuple the output gets zero padded and we can't use a Fill unless it is all zeros too
+    # Note, when dims is a tuple the output gets zero padded and we can't use a Fill unless it is all zeros
     allvals[] !== zero(T) && sum(catdims) > 1 && return Base._cat_t(dims, T, fs...)
 
     shape = Base.cat_shape(catdims, map(Base.cat_size, fs)::Tuple{Vararg{Union{Int,Dims}}})::Dims
@@ -14,3 +14,14 @@ end
 
 Base.vcat(vs::Fill...) = cat(vs...;dims=Val(1))
 Base.hcat(vs::Fill...) = cat(vs...;dims=Val(2))
+
+
+function Base.cat_t(::Type{T}, fs::Zeros...; dims) where T
+    catdims = Base.dims2cat(dims)
+    shape = Base.cat_shape(catdims, map(Base.cat_size, fs)::Tuple{Vararg{Union{Int,Dims}}})::Dims
+    return Zeros{T}(shape)
+end
+
+Base.vcat(vs::Zeros...) = cat(vs...;dims=Val(1))
+Base.hcat(vs::Zeros...) = cat(vs...;dims=Val(2))
+
--- a/src/fillcat.jl
+++ b/src/fillcat.jl
@@ -9,7 +9,7 @@ function Base.cat_t(::Type{T}, fs::Fill...; dims) where T
     # There might be some cases when it does not get padded which are not considered here
     allvals[] !== zero(T) && sum(catdims) > 1 && return Base._cat_t(dims, T, fs...)
 
-    shape = Base.cat_shape(catdims, map(Base.cat_size, fs)::Tuple{Vararg{Union{Int,Dims}}})::Dims
+    shape = cat_shape_fill(catdims, fs)
     return Fill(convert(T, fs[1].value), shape)
 end
 
@@ -19,7 +19,7 @@ Base.hcat(vs::Fill...) = cat(vs...;dims=Val(2))
 
 function Base.cat_t(::Type{T}, fs::Zeros...; dims) where T
     catdims = Base.dims2cat(dims)
-    shape = Base.cat_shape(catdims, map(Base.cat_size, fs)::Tuple{Vararg{Union{Int,Dims}}})::Dims
+    shape = cat_shape_fill(catdims, fs)
     return Zeros{T}(shape)
 end
 
@@ -34,10 +34,16 @@ function Base.cat_t(::Type{T}, fs::Ones...; dims) where T
     # There might be some cases when it does not get padded which are not considered here
     sum(catdims) > 1 && return Base._cat_t(dims, T, fs...)
 
-    shape = Base.cat_shape(catdims, map(Base.cat_size, fs)::Tuple{Vararg{Union{Int,Dims}}})::Dims
+    shape = cat_shape_fill(catdims, fs)
     return Ones{T}(shape)
 end
 
 Base.vcat(vs::Ones...) = cat(vs...;dims=Val(1))
 Base.hcat(vs::Ones...) = cat(vs...;dims=Val(2))
 
+
+if VERSION < v"1.6-"
+    cat_shape_fill(catdims, fs) = Base.cat_shape(catdims, (), map(Base.cat_size, fs)...)
+else
+    cat_shape_fill(catdims, fs) = Base.cat_shape(catdims, map(Base.cat_size, fs)::Tuple{Vararg{Union{Int,Dims}}})::Dims
+end
--- a/src/fillcat.jl
+++ b/src/fillcat.jl
@@ -7,7 +7,11 @@ function Base.cat_t(::Type{T}, fs::Fill...; dims) where T
 
     # When dims is a tuple the output gets zero padded and we can't use a Fill unless it is all zeros
     # There might be some cases when it does not get padded which are not considered here
-    allvals[] !== zero(T) && sum(catdims) > 1 && return Base._cat_t(dims, T, fs...)
+    
+    if sum(catdims) > 1 
+        allvals[] isa Number || return Base._cat_t(dims, T, fs...)
+        allvals[] !== zero(T) && return Base._cat_t(dims, T, fs...)
+    end
 
     shape = cat_shape_fill(catdims, fs)
     return Fill(convert(T, fs[1].value), shape)
--- a/src/fillcat.jl
+++ b/src/fillcat.jl
@@ -5,7 +5,8 @@ function Base.cat_t(::Type{T}, fs::Fill...; dims) where T
 
     catdims = Base.dims2cat(dims)
 
-    # Note, when dims is a tuple the output gets zero padded and we can't use a Fill unless it is all zeros
+    # When dims is a tuple the output gets zero padded and we can't use a Fill unless it is all zeros
+    # There might be some cases when it does not get padded which are not considered here
     allvals[] !== zero(T) && sum(catdims) > 1 && return Base._cat_t(dims, T, fs...)
 
     shape = Base.cat_shape(catdims, map(Base.cat_size, fs)::Tuple{Vararg{Union{Int,Dims}}})::Dims
@@ -24,4 +25,19 @@ end
 
 Base.vcat(vs::Zeros...) = cat(vs...;dims=Val(1))
 Base.hcat(vs::Zeros...) = cat(vs...;dims=Val(2))
+
+
+function Base.cat_t(::Type{T}, fs::Ones...; dims) where T
+    catdims = Base.dims2cat(dims)
+
+    # When dims is a tuple the output gets zero padded so we can't return a Ones
+    # There might be some cases when it does not get padded which are not considered here
+    sum(catdims) > 1 && return Base._cat_t(dims, T, fs...)
+
+    shape = Base.cat_shape(catdims, map(Base.cat_size, fs)::Tuple{Vararg{Union{Int,Dims}}})::Dims
+    return Ones{T}(shape)
+end
+
+Base.vcat(vs::Ones...) = cat(vs...;dims=Val(1))
+Base.hcat(vs::Ones...) = cat(vs...;dims=Val(2))
 
--- a/src/fillcat.jl
+++ b/src/fillcat.jl
@@ -1,0 +1,16 @@
+
+function Base.cat_t(::Type{T}, fs::Fill...; dims) where T
+    allvals = unique([f.value for f in fs])
+    length(allvals) > 1 && return Base._cat_t(dims, T,  fs...)
+
+    catdims = Base.dims2cat(dims)
+
+    # Note, when dims is a tuple the output gets zero padded and we can't use a Fill unless it is all zeros too
+    allvals[] !== zero(T) && sum(catdims) > 1 && return Base._cat_t(dims, T, fs...)
+
+    shape = Base.cat_shape(catdims, map(Base.cat_size, fs)::Tuple{Vararg{Union{Int,Dims}}})::Dims
+    return Fill(convert(T, fs[1].value), shape)
+end
+
+Base.vcat(vs::Fill...) = cat(vs...;dims=Val(1))
+Base.hcat(vs::Fill...) = cat(vs...;dims=Val(2))
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1259,7 +1259,8 @@ end
             @testset "Dim $dims" for dims in (
                 (1,2),
                 (2,3),
-                (1,3,4)
+                (1,3,4),
+                Iterators.take(3:5, 2)
             )
                 # This inserts a bunch of zeros so we can no longer assume the answer is a Fill
                 @test cat(Fill(1, s), Fill(1, s); dims=dims) == cat(fill(1, s), fill(1,s); dims=dims)
@@ -1300,7 +1301,8 @@ end
                 2,
                 Val(3),
                 (1,2),
-                (2,3)
+                (2,3),
+                Iterators.take(3:5, 2)
             )
                 res = cat(Zeros(s), Zeros(s); dims=dims) 
                 @test res isa Zeros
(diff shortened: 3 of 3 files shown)


dlfivefifty:
Won’t this not be type stable?

DrChainsaw:
> Won’t this not be type stable?

The fact that it might return either an Array or a Fill/Ones depending on shape or contents? 

Yes, I suppose so (ugh for double negation :), it will not be type stable). Is there any other option besides giving up on the idea?

DrChainsaw:
Maybe this was not your point, but performance better due to fewer allocations regardless of any additional type instability. code_warntype gives me basically idential output, but maybe I'm not using it at the right level. 

If this is about surprising the end user then I agree it's iffy and maybe is a reason why it should not happen.

```julia
julia> @benchmark cat(Fill(1,1), Fill(1,1);dims=1)
BenchmarkTools.Trial: 
  memory estimate:  864 bytes
  allocs estimate:  12
  --------------
  minimum time:     562.500 ns (0.00% GC)
  median time:      656.522 ns (0.00% GC)
  mean time:        893.306 ns (8.02% GC)
  maximum time:     26.051 μs (97.09% GC)
  --------------
  samples:          10000
  evals/sample:     184

julia> @benchmark cat(fill(1,1), fill(1,1);dims=1)
BenchmarkTools.Trial: 
  memory estimate:  704 bytes
  allocs estimate:  20
  --------------
  minimum time:     1.090 μs (0.00% GC)
  median time:      1.320 μs (0.00% GC)
  mean time:        1.610 μs (2.24% GC)
  maximum time:     363.690 μs (99.05% GC)
  --------------
  samples:          10000
  evals/sample:     10

julia> @benchmark cat(Fill(1,10000), Fill(1,10000);dims=1)
BenchmarkTools.Trial: 
  memory estimate:  864 bytes
  allocs estimate:  12
  --------------
  minimum time:     559.239 ns (0.00% GC)
  median time:      630.978 ns (0.00% GC)
  mean time:        861.916 ns (8.70% GC)
  maximum time:     26.658 μs (97.32% GC)
  --------------
  samples:          10000
  evals/sample:     184

julia> @benchmark cat(fill(1,10000), fill(1,10000);dims=1)
BenchmarkTools.Trial: 
  memory estimate:  313.14 KiB
  allocs estimate:  23
  --------------
  minimum time:     16.200 μs (0.00% GC)
  median time:      31.800 μs (0.00% GC)
  mean time:        76.410 μs (15.06% GC)
  maximum time:     3.978 ms (96.82% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

dlfivefifty:
The issue with type-unstable calls is not a single call with large arrays, but that the performance penalty propagates to other calls, which can have substantial (100x or more) performance degradations when called with small arrays. So I'm very much against introducing type-unstable code.

The `Ones` and `Zeros` cases should be fine, though please add `@inferred` to the tests.

What's your use case? If we do https://github.com/JuliaArrays/FillArrays.jl/issues/104 then the value can be known at compile for more general arrays. Though that PR isn't completely thought through, e.g., `Ones{BigFloat}` cannot become `SFill{one(BigFloat),BigFloat}` since `BigFloat` is not a bitstype:
```julia
julia> struct Foo{T} end

julia> Foo{1.0}() # fine since `1.0` is a bits type
Foo{1.0}()

julia> Foo{big(1.0)}()
ERROR: TypeError: in Type, in parameter, expected Int64, got a value of type BigFloat
Stacktrace:
 [1] top-level scope
   @ REPL[3]:1
```

DrChainsaw:
Hi,

As for my usecase, without making things too complicated; I'm using FillArrays to help keep memory utilization down, but some parts of my code concatenates arrays. I can easily work around in my code to just catch the Fills and give them special treatment, but then I saw #23 and thought I should just make a contribution instead as the problem looked simple on the surface.

I'm sorry if this is more a burden than a help. Feel free to close the issue if you don't want to waste more time on me. No hard feelings I promise :)

I also made the Ones and Zeros individual commits, so it should be possible to cherry pick them. 

#104 would certainly help, but there is still the padding case when concatenating along multiple dimensions which means that only Zeros can be type stable for all allowed inputs to `cat`. 

Here is an example of the "padding case" in case its not clear what I'm talking about:

```julia
julia> cat(ones(2), ones(2) ;dims=(1, 2))
4×2 Matrix{Float64}:
 1.0  0.0
 1.0  0.0
 0.0  1.0
 0.0  1.0
```
I guess problem is that I have 4 elements and I want 8. `Base` uses `zero(T)` to fill in the rest of the parameters and I think it is important to be consistent with that.

There is a lower level method which dispatches on the `dims` and maybe that can be used, but I'm not sure if it will help and its difficult to test as `cat` does not seem to be type stable to begin with, see below.

Another way I reasoned about the type stability is that for mutable arrays, all flavours of `cat` will allocate so if you put `cat` in a hot loop I don't think that type instability is your worst problem.

I understand about the small arrays, and I did include them in the benchmark. I understand about the propagation and that is ofc harder to benchmark, but will it ever create worse problems than the allocations? 

Here are benchmarks for small arrays with escaped input along with code_warntype. Note the difference between `Fill` (with capital `F`) and `fill`:

```julia

julia> @benchmark cat($(Fill(1,1)), $(Fill(1,1));dims=1)
BenchmarkTools.Trial: 
  memory estimate:  928 bytes
  allocs estimate:  14
  --------------
  minimum time:     582.558 ns (0.00% GC)
  median time:      697.674 ns (0.00% GC)
  mean time:        937.771 ns (8.54% GC)
  maximum time:     29.799 μs (94.53% GC)
  --------------
  samples:          10000
  evals/sample:     172

julia> @benchmark cat($(fill(1,1)), $(fill(1,1));dims=1)
BenchmarkTools.Trial: 
  memory estimate:  512 bytes
  allocs estimate:  18
  --------------
  minimum time:     969.231 ns (0.00% GC) 
  median time:      1.169 μs (0.00% GC)   
  mean time:        1.385 μs (1.92% GC)   
  maximum time:     268.692 μs (98.86% GC)
  --------------
  samples:          10000
  evals/sample:     13


julia> @code_warntype cat(Fill(1,1), Fill(1,1);dims=1)
Variables
  #unused#::Core.Const(Base.var"#cat##kw"())
  @_2::NamedTuple{(:dims,), Tuple{Int64}}
  @_3::Core.Const(cat)
  A::Tuple{Fill{Int64, 1, Tuple{Base.OneTo{Int64}}}, Fill{Int64, 1, Tuple{Base.OneTo{Int64}}}}
  dims::Int64
  @_6::Int64

Body::Any
1 ─ %1  = Base.haskey(@_2, :dims)::Core.Const(true)
│         %1
│         (@_6 = Base.getindex(@_2, :dims))
└──       goto #3
2 ─       Core.Const(:(Core.UndefKeywordError(:dims)))
└──       Core.Const(:(@_6 = Core.throw(%5)))
3 ┄       (dims = @_6)
│   %8  = (:dims,)::Core.Const((:dims,))
│   %9  = Core.apply_type(Core.NamedTuple, %8)::Core.Const(NamedTuple{(:dims,), T} where T<:Tuple)
│   %10 = Base.structdiff(@_2, %9)::Core.Const(NamedTuple())
│   %11 = Base.pairs(%10)::Core.Const(Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}())
│   %12 = Base.isempty(%11)::Core.Const(true)
│         %12
└──       goto #5
4 ─       Core.Const(:(Core.tuple(@_2, @_3)))
└──       Core.Const(:(Core._apply_iterate(Base.iterate, Base.kwerr, %15, A)))
5 ┄ %17 = Core.tuple(dims, @_3)::Core.PartialStruct(Tuple{Int64, typeof(cat)}, Any[Int64, Core.Const(cat)])
│   %18 = Core._apply_iterate(Base.iterate, Base.:(var"#cat#131"), %17, A)::Any
└──       return %18

julia> @code_warntype cat(fill(1,1), fill(1,1);dims=1)
Variables
  #unused#::Core.Const(Base.var"#cat##kw"())       
  @_2::NamedTuple{(:dims,), Tuple{Int64}}
  @_3::Core.Const(cat)
  A::Tuple{Vector{Int64}, Vector{Int64}}
  dims::Int64
  @_6::Int64

Body::Any
1 ─ %1  = Base.haskey(@_2, :dims)::Core.Const(true)
│         %1
│         (@_6 = Base.getindex(@_2, :dims))
└──       goto #3
2 ─       Core.Const(:(Core.UndefKeywordError(:dims)))
└──       Core.Const(:(@_6 = Core.throw(%5)))
3 ┄       (dims = @_6)
│   %8  = (:dims,)::Core.Const((:dims,))
│   %9  = Core.apply_type(Core.NamedTuple, %8)::Core.Const(NamedTuple{(:dims,), T} where T<:Tuple)
│   %10 = Base.structdiff(@_2, %9)::Core.Const(NamedTuple())
│   %11 = Base.pairs(%10)::Core.Const(Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}())
│   %12 = Base.isempty(%11)::Core.Const(true)
│         %12
└──       goto #5
4 ─       Core.Const(:(Core.tuple(@_2, @_3)))
└──       Core.Const(:(Core._apply_iterate(Base.iterate, Base.kwerr, %15, A)))
5 ┄ %17 = Core.tuple(dims, @_3)::Core.PartialStruct(Tuple{Int64, typeof(cat)}, Any[Int64, Core.Const(cat)])
│   %18 = Core._apply_iterate(Base.iterate, Base.:(var"#cat#131"), %17, A)::Any
└──       return %18
``` 

Maybe the type instability of `cat` for normal arrays is just a 1.6 bug?

dlfivefifty:
The padding `cat` is pretty specialised so I'm more worried about the type stability of `vcat(Fill(1,5), Fill(2,6))`. 

My recommendation is in your own code do something like:
```julia
_cat(a...; kwds...) = Base.cat(a...; kwds...)
_cat(a::AbstractFill...; kwds...) = # special case
```
Though you might be better off using `LazyArrays.Vcat`, etc., to always have a lazy concatenation.

DrChainsaw:
Even this case does not appear to be any more type-unstable than the normal `cat`:

```julia
julia> A,B = Fill(1, 1), Fill(2,1)
       @code_warntype cat(A,B; dims=1)
Variables
  #unused#::Core.Const(Base.var"#cat##kw"())
  @_2::NamedTuple{(:dims,), Tuple{Int64}}
  @_3::Core.Const(cat)
  A::Tuple{Fill{Int64, 1, Tuple{Base.OneTo{Int64}}}, Fill{Int64, 1, Tuple{Base.OneTo{Int64}}}}
  dims::Int64
  @_6::Int64

Body::Any
1 ─ %1  = Base.haskey(@_2, :dims)::Core.Const(true)
│         %1
│         (@_6 = Base.getindex(@_2, :dims))
└──       goto #3
2 ─       Core.Const(:(Core.UndefKeywordError(:dims)))
└──       Core.Const(:(@_6 = Core.throw(%5)))
3 ┄       (dims = @_6)
│   %8  = (:dims,)::Core.Const((:dims,))
│   %9  = Core.apply_type(Core.NamedTuple, %8)::Core.Const(NamedTuple{(:dims,), T} where T<:Tuple)
│   %10 = Base.structdiff(@_2, %9)::Core.Const(NamedTuple())
│   %11 = Base.pairs(%10)::Core.Const(Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}())
│   %12 = Base.isempty(%11)::Core.Const(true)
│         %12
└──       goto #5
4 ─       Core.Const(:(Core.tuple(@_2, @_3)))
└──       Core.Const(:(Core._apply_iterate(Base.iterate, Base.kwerr, %15, A)))
5 ┄ %17 = Core.tuple(dims, @_3)::Core.PartialStruct(Tuple{Int64, typeof(cat)}, Any[Int64, Core.Const(cat)])
│   %18 = Core._apply_iterate(Base.iterate, Base.:(var"#cat#131"), %17, A)::Any
└──       return %18

julia> a,b = fill(1, 1), fill(2,1)
       @code_warntype cat(a,b; dims=1)
Variables
  #unused#::Core.Const(Base.var"#cat##kw"())
  @_2::NamedTuple{(:dims,), Tuple{Int64}}
  @_3::Core.Const(cat)
  A::Tuple{Vector{Int64}, Vector{Int64}}
  dims::Int64
  @_6::Int64

Body::Any
1 ─ %1  = Base.haskey(@_2, :dims)::Core.Const(true)
│         %1
│         (@_6 = Base.getindex(@_2, :dims))
└──       goto #3
2 ─       Core.Const(:(Core.UndefKeywordError(:dims)))
└──       Core.Const(:(@_6 = Core.throw(%5)))
3 ┄       (dims = @_6)
│   %8  = (:dims,)::Core.Const((:dims,))
│   %9  = Core.apply_type(Core.NamedTuple, %8)::Core.Const(NamedTuple{(:dims,), T} where T<:Tuple)
│   %10 = Base.structdiff(@_2, %9)::Core.Const(NamedTuple())
│   %11 = Base.pairs(%10)::Core.Const(Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}())
│   %12 = Base.isempty(%11)::Core.Const(true)
│         %12
└──       goto #5
4 ─       Core.Const(:(Core.tuple(@_2, @_3)))
└──       Core.Const(:(Core._apply_iterate(Base.iterate, Base.kwerr, %15, A)))
5 ┄ %17 = Core.tuple(dims, @_3)::Core.PartialStruct(Tuple{Int64, typeof(cat)}, Any[Int64, Core.Const(cat)])
│   %18 = Core._apply_iterate(Base.iterate, Base.:(var"#cat#131"), %17, A)::Any
└──       return %18
```
Or am I doing this wrong?

There is some performance overhead for that case though, but I assume that it is because current implementation does a bit of double work in that case:

```julia
julia> A,B = Fill(1, 1), Fill(2,1)
       @benchmark cat($A,$B; dims=1)
BenchmarkTools.Trial: 
  memory estimate:  1.53 KiB
  allocs estimate:  34
  --------------
  minimum time:     1.330 μs (0.00% GC)
  median time:      1.610 μs (0.00% GC)
  mean time:        2.016 μs (6.58% GC)
  maximum time:     490.140 μs (98.96% GC)
  --------------
  samples:          10000
  evals/sample:     10

julia> a,b = fill(1, 1), fill(2,1)
       @benchmark cat($a,$b; dims=1)
BenchmarkTools.Trial: 
  memory estimate:  512 bytes
  allocs estimate:  18
  --------------
  minimum time:     1.030 μs (0.00% GC)
  median time:      1.260 μs (0.00% GC)
  mean time:        1.466 μs (2.48% GC)
  maximum time:     366.170 μs (99.18% GC)
  --------------
  samples:          10000
  evals/sample:     10
```
I think that some of the double work can be worked away at the cost of duplicating some functionality in Base.

Anyways, your concerns as well as me being overall nervous about tampering with such a fundamental thing is enough to deter me. Sorry for the noise.

dlfivefifty:
Can we leave this open? Someone else might be interested. In any case we should have the `Ones` and `Zeros` special cases

DrChainsaw:
Sure, sorry for being too trigger happy.

If there are any changes you'd like me to do I'm happy to incorporate them.

oschulz:
Thanks for this @DrChainsaw , this would be super-helpful for me (I often have to vcat vectors of constant/static statistical weights).

putianyi889:
general cat where values can be different can be achieved by [LazyArrays.jl](https://github.com/JuliaArrays/LazyArrays.jl).

oschulz:
> general cat where values can be different can be achieved by LazyArrays.jl.

Yes, I think the result would be very different from what @DrChainsaw is trying to achieve here.

I would very much prefer a type-stable `vcat` though, otherwise we may introduce type-instability in user applications in a way that's hard for them to escape (they'd need to add special handling for `FillArray`). I think specializing `vcat` and friends for `Ones` and `Zeros` would be very valuable though.

DrChainsaw:
Thanks @putianyi889. I do indeed have use for the functionality LazyArrays offers since my usecase also includes normal arrays (although I agree it does something very different from what FillArrays could do). 

However, I have tried to use [RecursiveArrayTools.jl](https://github.com/SciML/RecursiveArrayTools.jl) which seems to do a similar thing and the result was that I ended up with unfeasibly long compile times, so it seems that my use case would actually prefer a type-unstable version.

As for type-stable versions: I think the commit for `Zeros` from this PR can be used as is (except maybe it overloads a `Base` internal method), while `Ones` might need a separate version which dispatches directly on `vcat`/`hcat` due to [this](https://github.com/JuliaArrays/FillArrays.jl/pull/140#issuecomment-783287197). Maybe this inconsistency is also enough reason to only do `vcat` and `hcat` for `Zeros` too. I think the best course of action is to just close this and open a new PR. I'm not sure if people are waiting for me to do this though...

oschulz:
> I'm not sure if people are waiting for me to do this though...

Can't speak for other people, but I have actual statistics uses cases (merging sets of samples that all have weight one by construction) that would profit custom `vcat` on `Ones`.

oschulz:
> At least cat of Ones/Zeros only can be type stable. General cases can be left for now.

I'd vote for that, then. :-)

PR #144, opened 2021-05-05
Short form printing

mcabbott:
This adds a shorter printing for when a FillArray appears inside a tuple, or in an array.

Before:
```
julia> (1, Fill(ℯ,2), 3, Fill(π,4), 5)
(1, 2-element Fill{Irrational{:ℯ}}: entries equal to ℯ, 3, 4-element Fill{Irrational{:π}}: entries equal to π, 5)

julia> [Fill(i==j,2,2) for i in 1:3, j in 1:4]
3×4 Matrix{Fill{Bool, 2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}}:
 2×2 Fill{Bool}: entries equal to true   2×2 Fill{Bool}: entries equal to false  2×2 Fill{Bool}: entries equal to false  2×2 Fill{Bool}: entries equal to false
 2×2 Fill{Bool}: entries equal to false  2×2 Fill{Bool}: entries equal to true   2×2 Fill{Bool}: entries equal to false  2×2 Fill{Bool}: entries equal to false
 2×2 Fill{Bool}: entries equal to false  2×2 Fill{Bool}: entries equal to false  2×2 Fill{Bool}: entries equal to true   2×2 Fill{Bool}: entries equal to false
 
julia> Fill(99,1)
1-element Fill{Int64}: entries equal to 99

julia> Fill(1.0, (3:4,))
2-element Fill{Float64, 1, Tuple{UnitRange{Int64}}} with indices 3:4: entries equal to 1.0
```
After:
```
julia> (1, Fill(ℯ,2), 3, Fill(π,4), 5)
(1, Fill(ℯ, 2), 3, Fill(π, 4), 5)

julia> [Fill(i==j,2,2) for i in 1:3, j in 1:4]
3×4 Matrix{Fill{Bool, 2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}}:
 Fill(true, 2, 2)   Fill(false, 2, 2)  Fill(false, 2, 2)  Fill(false, 2, 2)
 Fill(false, 2, 2)  Fill(true, 2, 2)   Fill(false, 2, 2)  Fill(false, 2, 2)
 Fill(false, 2, 2)  Fill(false, 2, 2)  Fill(true, 2, 2)   Fill(false, 2, 2)

julia> Fill(99,1)
1-element Fill{Int64}, with entry equal to 99

julia> Fill(1.0, (3:4,))
[...]

PR #146, opened 2021-05-15
`map` and `mapreduce`

mcabbott:
This should make `map(f, A, B, C)`, and `mapreduce(f, op, A, B, C)` work for FillArrays.

One motivation is that Base's `mapreduce` isn't fast for multiple arrays, but if we peel off the ones which are `Fill` then we can go the fast path. The other is that a mix of CuArrays & Fill seems to give errors:
```
julia> mapreduce((x,y) -> x*conj(y), +, Fill(1.0, 3), cu(ones(3)))
ERROR: scalar getindex is disallowed
```
Edit -- scope has crept a little, it now also handles `reduce`, which has this result:
```
julia> sum(Fill(3,4,5), dims=1)
1×5 Matrix{Int64}:  # master
 12  12  12  12  12

1×5 Fill{Int64}: entries equal to 12  # this PR
```
Happy to remove / separate that if you prefer.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -782,9 +782,12 @@ end
 
     @test mapreduce(exp, +, Y) == mapreduce(exp, +, y)
     @test mapreduce(exp, +, Y; dims=2) == mapreduce(exp, +, y; dims=2)
-    @test mapreduce(exp, +, Y; dims=(1,), init=5.0) == mapreduce(exp, +, y; dims=(1,), init=5.0)
     @test mapreduce(identity, +, Y) == sum(y) == sum(Y)
     @test mapreduce(identity, +, Y, dims=1) == sum(y, dims=1) == sum(Y, dims=1)
+
+    if VERSION >= v"1.5"
+        @test mapreduce(exp, +, Y; dims=(1,), init=5.0) == mapreduce(exp, +, y; dims=(1,), init=5.0)
+    end
 
     # Two arrays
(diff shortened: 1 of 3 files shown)


test/runtests.jl
```diff
@@ -758,15 +758,46 @@ end
...
+    @test_throws DimensionMismatch map(+, x2', x2)
+end
+
+@testset "mapreduce" begin
+    x = rand(3, 4)
+    y = fill(1.0, 3, 4)
+    Y = Fill(1.0, 3, 4)
+    O = Ones(3, 4)
+
+    @test_broken mapreduce(exp, +, Y) == mapreduce(exp, +, y)
```
mcabbott, line 782: This case leads to a stack overflow, I guess it's just this:
```
julia> reduce(+, Fill(3,4))
ERROR: StackOverflowError:
```

mcabbott: I've removed that method for now.

mcabbott:
Error on 1.0 was real, "UndefVarError: _InitialValue not defined".

I've just disabled these methods on old versions, rather than try to figure out how 1.0 did this & exactly when the change happened. Without them you will more often get an Array, as you do without this PR.

mcabbott:
Oh right, sorry, looks like vararg `mapreduce` didn't exist at all then. Will omit those bits.

mcabbott:
Bump?

dlfivefifty:
Great if you bump the version number I can tag a new release

mcabbott:
Can do, or else #144 has a version bump, too.

Issue #148, opened 2021-06-26
getindex of Ones with Trues


JeffFessler:
I managed to hit a simple looking case that probably needs another `getindex` variation:

```julia
using FillArrays

x = Ones{Float32}(4,5) # 2D
b = Trues(length(x))
x[b] # works fine

y = Ones{Float32}(3) # 1D
c = Trues(length(y))
y[c] # fails


ERROR: LoadError: MethodError: getindex(::Ones{Float32, 1, Tuple{Base.OneTo{Int64}}}, ::Trues{1, Tuple{Base.OneTo{Int64}}}) is ambiguous. Candidates:
  getindex(x::AbstractArray{T, N}, mask::Ones{Bool, N, Tuple{Vararg{Base.OneTo{Int64}, N}}}) where {T, N} in FillArrays at /Users/fessler/.julia/packages/FillArrays/cVkp8/src/trues.jl:39
  getindex(A::FillArrays.AbstractFill, kr::AbstractVector{Bool}) in FillArrays at /Users/fessler/.julia/packages/FillArrays/cVkp8/src/FillArrays.jl:183
  getindex(A::FillArrays.AbstractFill, kr::AbstractArray{Bool, N} where N) in FillArrays at /Users/fessler/.julia/packages/FillArrays/cVkp8/src/FillArrays.jl:184
Possible fix, define
  getindex(::FillArrays.AbstractFill{T, 1, Axes} where Axes, ::Trues{1, Tuple{Base.OneTo{Int64}}}) where T
```

Closest related issue or PR I could find was this one: https://github.com/JuliaArrays/FillArrays.jl/issues/33

PR #149, opened 2021-06-26
fix getindex of Fill with Trues

JeffFessler:
Hopefully addresses #148
Now with tests...

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FillArrays"
 uuid = "1a297f60-69ca-5386-bcde-b61e274b549b"
-version = "0.11.7"
+version = "0.12.0"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FillArrays"
 uuid = "1a297f60-69ca-5386-bcde-b61e274b549b"
-version = "0.12.0"
+version = "0.11.8"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/trues.jl
+++ b/src/trues.jl
@@ -46,8 +46,8 @@ end
 
 # https://github.com/JuliaArrays/FillArrays.jl/issues/148
 function Base.getindex(
-    a::FillArrays.AbstractFill{T, 1, Axes} where Axes, 
-    b::Trues{1, Tuple{Base.OneTo{Int64}}}) where T
+    a::AbstractFill{T, 1, Axes} where Axes,
+    b::Trues{1, Tuple{Base.OneTo{Int}}}) where T
     @boundscheck length(a) == length(b) || throw(BoundsError(a, b))
-    return Fill(a[1], length(a))
+    return Fill(getindex_value(a), length(a))
 end
--- a/src/trues.jl
+++ b/src/trues.jl
@@ -43,3 +43,11 @@ function Base.getindex(x::AbstractArray{T,N}, mask::Trues{N, NTuple{N,Base.OneTo
     end
     return x[trues(size(x))] # else revert to usual getindex method
 end
+
+# https://github.com/JuliaArrays/FillArrays.jl/issues/148
+function Base.getindex(
+    a::FillArrays.AbstractFill{T, 1, Axes} where Axes, 
+    b::Trues{1, Tuple{Base.OneTo{Int64}}}) where T
+    @boundscheck length(a) == length(b) || throw(BoundsError(a, b))
+    return Fill(a[1], length(a))
+end
--- a/src/trues.jl
+++ b/src/trues.jl
@@ -46,7 +46,7 @@ end
 
 # https://github.com/JuliaArrays/FillArrays.jl/issues/148
 function Base.getindex(
-    a::AbstractFill{T, 1, Axes} where Axes,
+    a::AbstractFill{T, 1, Tuple{Base.OneTo{Int}}},
     b::Trues{1, Tuple{Base.OneTo{Int}}}) where T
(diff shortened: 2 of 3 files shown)


src/trues.jl
```diff
@@ -43,3 +43,11 @@ function Base.getindex(x::AbstractArray{T,N}, mask::Trues{N, NTuple{N,Base.OneTo
     end
     return x[trues(size(x))] # else revert to usual getindex method
 end
+
+# https://github.com/JuliaArrays/FillArrays.jl/issues/148
+function Base.getindex(
+    a::FillArrays.AbstractFill{T, 1, Axes} where Axes, 
+    b::Trues{1, Tuple{Base.OneTo{Int64}}}) where T
```
dlfivefifty, line 50: ` Int64 ` -> ` Int `

src/trues.jl
```diff
@@ -43,3 +43,11 @@ function Base.getindex(x::AbstractArray{T,N}, mask::Trues{N, NTuple{N,Base.OneTo
     end
     return x[trues(size(x))] # else revert to usual getindex method
 end
+
+# https://github.com/JuliaArrays/FillArrays.jl/issues/148
+function Base.getindex(
+    a::FillArrays.AbstractFill{T, 1, Axes} where Axes,
```
dlfivefifty, line 49: Remove ` FillArrays. `

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "FillArrays"
 uuid = "1a297f60-69ca-5386-bcde-b61e274b549b"
-version = "0.11.7"
+version = "0.12.0"
```
dlfivefifty, line 3: Just do a patch version ` 0.11.8 `

src/trues.jl
```diff
@@ -43,3 +43,11 @@ function Base.getindex(x::AbstractArray{T,N}, mask::Trues{N, NTuple{N,Base.OneTo
     end
     return x[trues(size(x))] # else revert to usual getindex method
 end
+
+# https://github.com/JuliaArrays/FillArrays.jl/issues/148
+function Base.getindex(
+    a::FillArrays.AbstractFill{T, 1, Axes} where Axes, 
+    b::Trues{1, Tuple{Base.OneTo{Int64}}}) where T
+    @boundscheck length(a) == length(b) || throw(BoundsError(a, b))
+    return Fill(a[1], length(a))
```
dlfivefifty, line 52: Use ` getindex_value(a) ` in case ` a ` is empty

JeffFessler:
Thanks for the quick feedback - all addressed hopefully.

src/trues.jl
```diff
@@ -43,3 +43,11 @@ function Base.getindex(x::AbstractArray{T,N}, mask::Trues{N, NTuple{N,Base.OneTo
     end
     return x[trues(size(x))] # else revert to usual getindex method
 end
+
+# https://github.com/JuliaArrays/FillArrays.jl/issues/148
+function Base.getindex(
+    a::AbstractFill{T, 1, Axes} where Axes,
+    b::Trues{1, Tuple{Base.OneTo{Int}}}) where T
+    @boundscheck length(a) == length(b) || throw(BoundsError(a, b))
+    return Fill(getindex_value(a), length(a))
```
dlfivefifty, line 52: Need to check ` axes ` match

test/runtests.jl
```diff
@@ -1204,6 +1204,8 @@ end
     @test_throws BoundsError ones(3)[Trues(2)]
     @test_throws BoundsError setindex!(ones(3), zeros(3), Trues(2))
     @test_throws DimensionMismatch setindex!(ones(2), zeros(3), Trues(2))
+    @test Ones(3)[Trues(3)] == Ones(3)
```
dlfivefifty, line 1207: Add test with offset axes

dlfivefifty (changes requested):

src/trues.jl
JeffFessler: Short answer:
Because `b` here is set to be `Tuple{Base.OneTo{Int}}}`, instead of adding check and a test, I have simply specialized the method so that it applies only to `a` with those same axes.  That will suffice for my use cases, and what to do more generally seems unclear.

You can ignore the rest of this comment if you are satisfied with that solution.
I am writing it both as notes for myself and in case it helps someone else generalize the method in the future if there is a need.

I am unsure that the `axes` must match when doing `getindex` with a 1D Bool array like we are here.

The following 1D logical indexing on RHS works fine in Base:
```julia
ones(2,3)[trues(2,3)] == ones(2,3)[trues(6)]
```

And similar 1D logical indexing works fine in `FillArrays` already:
```julia
Ones((3:5, 2:3))[trues(6)]
```

The axes do not even have the same dimensions in those two examples, so clearly the axes do not match, yet it works as expected.
In general when one does `getindex` with a 1D Bool vector, the axes become moot,
because of arbitrary sampling, e.g.,
`(2:4)[[true, false, true]]`
or even
`Ones(3)[[true, false, true]]`

On the other hand, this throws an error in FA currently:
`Ones((2:4,))[[true,false,true]]`

It is unclear to me from the docs whether throwing an error is the right thing to do or not.
https://docs.julialang.org/en/v1/devdocs/offset-arrays/
Resolving that (if needed) would be a separate issue from this PR.

test/runtests.jl
JeffFessler: With the restricted I added above, now something like
`Ones((2:4,))[Trues(3)]`
throws an error about method ambiguity, which seems appropriate.
I'm assuming we don't want a test for that.
The manual hints that there is an ambiguity for linear indexing
https://docs.julialang.org/en/v1/devdocs/offset-arrays/#Linear-indexing-(LinearIndices)
but does not seem to address the logical indexing case.

dlfivefifty: That’s fine, didn’t realise that things were so ill-defined

dlfivefifty (approved):

JeffFessler:
Can you merge and tag now?  It would help me downstream.  Thanks.

PR #152, opened 2021-07-18
Add methods for `mean`, `std`

mcabbott:
This solves the following problem:
```julia
julia> using FillArrays, Statistics

julia> mean(Fill(3,4,5))
3.0

julia> mean(Fill(3,4,5), dims=1)
ERROR: ArgumentError: Cannot setindex! to 3.0 for an AbstractFill with value 12.0.

julia> cor(Fill(3, 4))
1.0

julia> cor(Fill(3, 4, 5))
ERROR: ArgumentError: Cannot setindex! to 1.0 for an AbstractFill with value 0.0.
```
Adds dependence on one more standard lib, which I hope is OK.

### DIFF
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -601,14 +601,10 @@ function mean(f::Union{Function, Type}, A::AbstractFill; dims=(:))
         Fill(val, ntuple(d -> d in dims ? 1 : size(A,d), ndims(A))...)
 end
 
-for fun in (:std, :var)
-    @eval function $fun(A::AbstractFill{T}; corrected::Bool=true, mean=nothing, dims=(:)) where {T<:Number}
-        if mean !== nothing
-            mean ≈ Statistics.mean(A; dims=dims) || throw(ArgumentError("pre-computed mean is incorrect"))
-        end
-        dims isa Colon ? zero(float(T)) : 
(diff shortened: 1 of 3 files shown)


src/FillArrays.jl
```diff
@@ -589,6 +590,30 @@ function in(x, A::RectDiagonal{<:Number})
...
+function mean(f, A::AbstractFill; dims=(:))
+    val = float(f(getindex_value(A)))
+    dims isa Colon ? val : 
+        Fill(val, ntuple(d -> d in dims ? 1 : size(A,d), ndims(A)))
+end
+
+for fun in (:std, :var)
+    @eval function $fun(A::AbstractFill{T}; corrected::Bool=true, mean=nothing, dims=(:)) where {T<:Number}
+        if mean !== nothing
+            mean ≈ Statistics.mean(A; dims=dims) || throw(ArgumentError("pre-computed mean is incorrect"))
```
mcabbott, line 607: Maybe this should change, the usual method does let you give the wrong mean -- I was lazy to make it precisely reproduce the same (wrong) answer. Perhaps just ignoring it would also be OK?
```
julia> std(fill(1,2), mean=2)
1.4142135623730951

julia> std(fill(1,10), mean=2)
1.0540925533894598
```

dlfivefifty:
bump the patch version and I can tag

PR #154, opened 2021-07-27
Add note about broadcasting etc. to the readme

mcabbott:
A spare commit from some PR. Might be good to mention this.

It's possible that this should be upgraded to a more scary warning about purity assumptions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ julia> println.(Fill(pi, 10))
 10-element Fill{Nothing}: entries equal to nothing
 ```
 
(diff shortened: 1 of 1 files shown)


dlfivefifty:
You mean warning in the readme? Sure.

A warning in code when first run is a bad idea, as it's impossible to turn off warnings when wanted.

mcabbott:
Yes, I just mean documentation, not `@warn`. See what you think of the latest.

dlfivefifty:
Thanks!

PR #156, opened 2021-08-18
Methods for `issymmetric` & `ishermitian`

mcabbott:
Right now these iterate through, needlessly:
```
julia> @btime ishermitian(Fill(1, 10, 10))
  33.911 ns (0 allocations: 0 bytes)
true

julia> @btime ishermitian(Fill(1, 1000, 1000))
  215.875 μs (0 allocations: 0 bytes)
true
```

### DIFF
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -9,7 +9,8 @@ import Base: size, getindex, setindex!, IndexStyle, checkbounds, convert,
     show, view, in, mapreduce
 
 import LinearAlgebra: rank, svdvals!, tril, triu, tril!, triu!, diag, transpose, adjoint, fill!,
-    dot, norm2, norm1, normInf, normMinusInf, normp, lmul!, rmul!, diagzero, AbstractTriangular, AdjointAbsVec
+    dot, norm2, norm1, normInf, normMinusInf, normp, lmul!, rmul!, diagzero, AbstractTriangular, AdjointAbsVec,
+    issymmetric, ishermitian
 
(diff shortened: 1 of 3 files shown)


src/FillArrays.jl
```diff
@@ -61,6 +62,8 @@ end
 rank(F::AbstractFill) = iszero(getindex_value(F)) ? 0 : 1
 IndexStyle(::Type{<:AbstractFill{<:Any,N,<:NTuple{N,Base.OneTo{Int}}}}) where N = IndexLinear()
 
+issymmetric(F::AbstractFill) = axes(F,1) == axes(F,2)
```
dlfivefifty, line 65: Restrict to matrices?

mcabbott: Oh right, will do.

src/FillArrays.jl
```diff
@@ -61,6 +62,8 @@ end
 rank(F::AbstractFill) = iszero(getindex_value(F)) ? 0 : 1
 IndexStyle(::Type{<:AbstractFill{<:Any,N,<:NTuple{N,Base.OneTo{Int}}}}) where N = IndexLinear()
 
+issymmetric(F::AbstractFill{<:Any, 2}) = axes(F,1) == axes(F,2)
+ishermitian(F::AbstractFill{<:Any, 2}) = issymmetric(F) & iszero(imag(getindex_value(F)))
```
dlfivefifty, line 66: Better to use `&&`

PR #163, opened 2021-10-01
Solve all problems with extra indirection

### LINKED ISSUE #162: `Base.getindex(x::AbstractArray{T,N}, mask::Trues{N, NTuple{N,Base.OneTo{Int}}})` is downstream chaos
Found in the ambiguity detection in SciMLBase after a FillArrays.jl dependency was added:

https://github.com/SciML/RecursiveArrayTools.jl/runs/3752855248#step:6:202

The issue stems from `Base.getindex(x::AbstractArray{T,N}, mask::Trues{N, NTuple{N,Base.OneTo{Int}}})` types of definitions. The abstract array interface is defined as https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array . This has definitions `getindex(A, I...)` and `setindex!(A, X, I...)` specific to the array `A`, but possibly non-specific to the type `I`. To handle some ambiguities we distinguish over a decently sized set:

https://github.com/SciML/SciMLBase.jl/blob/master/src/solutions/solution_interface.jl#L31-L32

but adding FillArrays to the set still fails:

https://github.com/SciML/SciMLBase.jl/pull/105

I think downstream packages defining any overloads of the form `Base.getindex(x::AbstractArray{T,N}, I::T)` where `T` is the new package thing will cause ambiguity issues and require all downstream array types to define the intersection method `Base.getindex(x::A, I::T)`, which means the only way it can be handled correctly is for type package `T` to be a dependency of all downstream array packages `A` (😱). To me that doesn't sound like a tenable solution.

On top of that, I would presume that this definition, since it introduces ambiguities into all of these downstream "open indexing" definitions, it's also invalidating a ton of downstream `getindex` and `setindex!` operations as well. So it's likely not just an issue causing ambiguity errors but also hitting compile times.

Maybe @timholy or @mbauman have an opinion on whether these overloads are okay. I would think the right thing would be to remove them.

timholy:
The `getindex` and `setindex!` methods for `Trues` are limited
while also risking  ambiguities. This replaces those definitions
with a specialization for `to_indices` that avoids such problems.

Closes #162

### DIFF
--- a/src/trues.jl
+++ b/src/trues.jl
@@ -22,33 +22,8 @@ const Falses = Zeros{Bool, N, Axes} where {N, Axes}
 
 
 # y[mask] = x when mask isa Trues (cf y[:] = x)
-# Supported here only for arrays with standard OneTo axes.
-function Base.setindex!(y::AbstractArray{T,N}, x, mask::Trues{N, NTuple{N,Base.OneTo{Int}}}) where {T,N}
-    if axes(x) isa NTuple{N,Base.OneTo{Int}} && axes(y) isa NTuple{N,Base.OneTo{Int}}
-        @boundscheck size(y) == size(mask) || throw(BoundsError(y, mask))
-        @boundscheck size(x) == size(mask) || throw(DimensionMismatch(
-            "tried to assign $(length(x)) elements to $(length(y)) destinations"))
-        @boundscheck checkbounds(y, mask)
-        return copyto!(y, x)
-    end
-    return setindex!(y, x, trues(size(mask))) # fall back on usual setindex!
-end
 
-# x[mask] when mask isa Trues (cf x[trues(size(x))] or x[:])
-# Supported here only for arrays with standard OneTo axes.
-function Base.getindex(x::AbstractArray{T,N}, mask::Trues{N, NTuple{N,Base.OneTo{Int}}}) where {T,N}
-    if axes(x) isa NTuple{N,Base.OneTo{Int}}
-       @boundscheck size(x) == size(mask) || throw(BoundsError(x, mask))
-       return vec(x)
-    end
-    return x[trues(size(x))] # else revert to usual getindex method
-end
-
-# https://github.com/JuliaArrays/FillArrays.jl/issues/148 and 150
-function Base.getindex(
-    a::AbstractFill{T, N, Tuple{Vararg{Base.OneTo{Int}, N}}},
-    b::Trues{N, Tuple{Vararg{Base.OneTo{Int}, N}}},
-) where {T, N}
-    @boundscheck size(a) == size(b) || throw(BoundsError(a, b))
-    return Fill(getindex_value(a), length(a))
+function Base.to_indices(A::AbstractArray{T,N}, inds, I::Tuple{Trues{N}}) where {T,N}
+    @boundscheck axes(A) == axes(I[1]) || Base.throw_boundserror(A, I[1])
+    (vec(LinearIndices(A)),)
 end


ChrisRackauckas:
Boom. Thanks. I knew you'd be the one to know the solution haha

timholy:
The best answer to ambiguities is almost always to delete methods. After all, if you have 0 methods, it's guaranteed there will be 0 ambiguities :laughing:.

dlfivefifty:
Note the `to_indices` call is not actually invoked...

I think this is probably more proof that the tests are not testing the function call. But TBH I don't really understand the use case for these special `getindex` and `setindex!` overloads so I'm fine with deleting them (along with the `to_indices` overload you added).... The original author @JeffFessler can of course propose a better alternative with relevant tests.

timholy:
Here's a test that now passes that didn't before:
```julia
    mask = OffsetArray(Trues(dim), 0:3, -2:2)
    x = OffsetArray(randn(dim), axes(mask)...)
    @test x[mask] == vec(x)
    @test_throws BoundsError x[parent(mask)]
```
Unfortunately, simply loading OffsetArrays introduces an ambiguity with your reshape methods. If you delete all your support for `reshape` that goes away, and the only thing that fails are your `≡ Fill(2, 2, 3)`-type tests. There may be good ways to solve that too, but I haven't looked. The first question is: how important is it to avoid the `ReshapedArray` wrapper?

timholy:
> Note the to_indices call is not actually invoked...

This should fix it. I am guessing it's a performance optimization to save time in checking each boolean value.

dlfivefifty:
Removing `reshape` is a step too far, I think. Feels like a race to the bottom where we in the end remove all functionality.... and the argument can be made that OffsetArrays.jl should remove it's `reshape` overloads.

In the end ambiguities are just a fact of life in combining Julia packages, trying to avoid them completely is not realistic. This is where a glue package called OffsetFillArrays.jl or whatever would come in (with future Pkg support for making it mandatory when one uses the two packages together)

dlfivefifty:
> I am guessing it's a performance optimization to save time in checking each boolean value.

Yes of course, the question is when would anyone in practice call `x[Trues(...)]` in the first flace

timholy:
I bet there's a way around it. You're right, though, that Base has `reshape` methods that specialize both the array-type and the shape-types. That's usually a sign of poor design (fault: mine). See https://docs.julialang.org/en/v1/manual/methods/#man-method-design-ambiguities

dlfivefifty:
I wonder if long term it’s better to make an ordering of which comes first, array overloads or shapes. If so we could do
```Julia 
reshape(A…) = reshapebyshapes(A…)
reshapebyshapes(A…) = reshapebyarraytype(A…)
reshapebyarraytype(A…) = defaultreshape(A…)
```
Then if one only does overloads by
```Julia
reshapebyshapes(A, ::MyShape) = …
reshapebyarraytype(A::MyArray, sh…) = …
```
We would avoid ambiguity 


But would need a change in ` Base `

timholy:
Yep, it's a tricky design problem. No time to look into it now, but it turns out that when I tested reshape method deletion I also deleted all the support routines. If you keep those and just delete the `reshape` methods, everything passes. Moreover, this reduces the number of ambiguities in this package (even without loading OffsetArrays) from 56 to 52.

Shall I pile that on here or do you want a separate PR?

dlfivefifty:
Does it still return a `Fill ` ? I’m pretty sure I rely on this downstream (in particular ` permutedims ` which I believe calls ` reshape ` )

timholy:
Yes (the tests check as much)

JeffFessler:
> Yes of course, the question is when would anyone in practice call `x[Trues(...)]` in the first flace

My use case for `x[Trues(...)]` is in [ImageGeoms.jl](https://github.com/JuliaImageRecon/ImageGeoms.jl) and in fact the Literate docs for that package specifically refers to this specific feature of FillArrays: https://juliaimagerecon.github.io/ImageGeoms.jl/stable/examples/2-mask

So I very much hope that some version of `x[Trues(...)]` remains supported in the most efficient way possible.  The original version basically mapped `x[Trues(...)]` into `vec(x)` which (as you know) is *non-allocating* (akin to a `view`).  It looks to me like the proposed version will end up doing `A[vec(LinearIndices(A))]` which will be allocating.  This would significantly degrade performance if I am understanding correctly.

Is there some other way to restrict the types of this `getindex` method to avoid method ambiguities?  I would like to retain the memory efficiency of `vec(x)` somehow.

I need to think more about whether the proposed deletion of `setindex!` will degrade performance.  Can you give me a couple days to run some timing tests?

On a perhaps slightly related note, to support all the functionality needed efficiently I also had to write a `getindex!` method that (to my surprise) was not supported directly in Base:
https://github.com/JuliaImageRecon/ImageGeoms.jl/blob/ac86239b72518c6b920439d49f74fcf728707ab2/src/mask.jl#L17
That code does use `LinearIndices` but it is inherently non-allocating.

dlfivefifty:
@JeffFessler Yes no rush. It's possible to fix the issue without deleting everything.

Can you add tests so that we make sure we don't lose the functionality you need?

timholy:
Yes, that's a good idea. I really doubt there will be any performance implications but you never know unless you check. No rush here.

ChrisRackauckas:
Can we merge this? I think that's been ample time to test.

ChrisRackauckas:
and a tag?

dlfivefifty:
Sure, done. Note (I assume) you have admin rights and can just tag a new release yourself in the future

dlfivefifty:
Err no one bumped the version. If you want a PR tagged please up the version first.

Issue #170, opened 2022-02-01
`nxn Fill` addition `UniformScaling` errors


theogf:
Here is a MWE:
```julia
julia> K = Fill(1.0, 2, 2)
julia> K + 1e-6 * I
ERROR: ArgumentError: Cannot setindex! to 1.000001 for an AbstractFill with value 1.0.
Stacktrace:
 [1] setindex!
   @ ~/.julia/packages/FillArrays/Vzxer/src/FillArrays.jl:53 [inlined]
 [2] +(A::Fill{Float64, 2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}, J::UniformScaling{Float64})
   @ LinearAlgebra ~/julia-1.7.1/share/julia/stdlib/v1.7/LinearAlgebra/src/uniformscaling.jl:219
 [3] top-level scope
   @ REPL[28]:1
```
Observed at https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/432

dlfivefifty:
Always include the full stack trace as otherwise we can't see where the error is

theogf:
Updated! Matrix addition with a `UniformScaling` creates a new matrix of the same type of the matrix and adds the values in place.

dlfivefifty:
OK so probably we need to overload `+(::Fill, ::UniformScaling)` and varients

devmotion:
I opened a PR that fixes the issue: https://github.com/JuliaArrays/FillArrays.jl/pull/171

PR #171, opened 2022-02-02
BREAKING: Support addition and subtraction with `UniformScaling` + drop support for Julia < 1.6

devmotion:
Fixes https://github.com/JuliaArrays/FillArrays.jl/issues/170.

### DIFF
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -19,11 +19,7 @@ import Statistics: mean, std, var, cov, cor
 
 export Zeros, Ones, Fill, Eye, Trues, Falses
 
-if VERSION < v"1.6-"
-    oneto(n) = Base.OneTo(n)
-else
-    import Base: oneto
-end
+import Base: oneto
 
 """
     AbstractFill{T, N, Axes} <: AbstractArray{T, N}
@@ -522,10 +518,7 @@ end
 sum(x::AbstractFill) = getindex_value(x)*length(x)
 sum(x::Zeros) = getindex_value(x)
 
-# define `sum(::Callable, ::AbstractFill)` to avoid method ambiguity errors on Julia 1.0
-sum(f, x::AbstractFill) = _sum(f, x)
-sum(f::Base.Callable, x::AbstractFill) = _sum(f, x)
-_sum(f, x::AbstractFill) = length(x) * f(getindex_value(x))
+sum(f, x::AbstractFill) = length(x) * f(getindex_value(x))
 
 cumsum(x::AbstractFill{<:Any,1}) = range(getindex_value(x); step=getindex_value(x),
                                                     length=length(x))
@@ -695,16 +688,14 @@ function Base.show(io::IO, x::AbstractFill)  # for example (Fill(π,3),)
 end
 Base.show(io::IO, x::Eye) = print(io, "Eye(", size(x,1), ")")
 
-if VERSION ≥ v"1.5"
-    Base.array_summary(io::IO, ::Zeros{T}, inds::Tuple{Vararg{Base.OneTo}}) where T =
-        print(io, Base.dims2string(length.(inds)), " Zeros{$T}")
-    Base.array_summary(io::IO, ::Ones{T}, inds::Tuple{Vararg{Base.OneTo}}) where T =
-        print(io, Base.dims2string(length.(inds)), " Ones{$T}")
-    Base.array_summary(io::IO, a::Fill{T}, inds::Tuple{Vararg{Base.OneTo}}) where T =
-        print(io, Base.dims2string(length.(inds)), " Fill{$T}")
-    Base.array_summary(io::IO, a::Eye{T}, inds::Tuple{Vararg{Base.OneTo}}) where T =
-        print(io, Base.dims2string(length.(inds)), " Eye{$T}")
-end
+Base.array_summary(io::IO, ::Zeros{T}, inds::Tuple{Vararg{Base.OneTo}}) where T =
+    print(io, Base.dims2string(length.(inds)), " Zeros{$T}")
+Base.array_summary(io::IO, ::Ones{T}, inds::Tuple{Vararg{Base.OneTo}}) where T =
+    print(io, Base.dims2string(length.(inds)), " Ones{$T}")
+Base.array_summary(io::IO, a::Fill{T}, inds::Tuple{Vararg{Base.OneTo}}) where T =
+    print(io, Base.dims2string(length.(inds)), " Fill{$T}")
+Base.array_summary(io::IO, a::Eye{T}, inds::Tuple{Vararg{Base.OneTo}}) where T =
+    print(io, Base.dims2string(length.(inds)), " Eye{$T}")
 
 
 ##
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -207,7 +207,10 @@ function +(a::Fill{T, 1}, b::AbstractRange) where {T}
     return a.value .+ b
 end
 +(a::AbstractRange, b::AbstractFill) = b + a
-+(a::AbstractFill{<:Any,2}, b::UniformScaling) = a + Diagonal(Fill(b.λ,(axes(a,1),)))
+function +(a::AbstractFill{<:Any,2}, b::UniformScaling)
+    n = LinearAlgebra.checksquare(a)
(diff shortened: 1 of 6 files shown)


src/fillalgebra.jl
```diff
@@ -261,6 +261,33 @@ end
...
 
 
+function +(a::Zeros{T, 2}, b::UniformScaling{V}) where {T, V<:Number}
+    n = LinearAlgebra.checksquare(a)
+    return Diagonal(Fill{promote_type(T,V)}(b.λ, n))
+end
+function +(a::RectOrDiagonal{T, <:AbstractFill{T, 1}}, b::UniformScaling) where {T}
+    LinearAlgebra.checksquare(a)
+    return Diagonal(a.diag .+ b.λ)
+end
```
devmotion, line 271: The reverse order falls back to these definitions: https://github.com/JuliaLang/julia/blob/e3b681c6872d87d74db63dd7ec3da2beee8e3ae4/stdlib/LinearAlgebra/src/uniformscaling.jl#L153

src/FillArrays.jl
```diff
@@ -207,9 +207,11 @@ function +(a::Fill{T, 1}, b::AbstractRange) where {T}
     return a.value .+ b
 end
 +(a::AbstractRange, b::AbstractFill) = b + a
++(a::AbstractFill{<:Any,2}, b::UniformScaling) = Array(a) + b
```
devmotion, line 210: The reverse order `+(::UniformScaling, ::AbstractFill{<:Any,2})` falls back to this definition: https://github.com/JuliaLang/julia/blob/e3b681c6872d87d74db63dd7ec3da2beee8e3ae4/stdlib/LinearAlgebra/src/uniformscaling.jl#L153

src/FillArrays.jl
```diff
@@ -207,9 +207,11 @@ function +(a::Fill{T, 1}, b::AbstractRange) where {T}
...
 
 -(a::AbstractFill, b::AbstractRange) = a + (-b)
 -(a::AbstractRange, b::AbstractFill) = a + (-b)
+-(a::UniformScaling, b::AbstractFill{<:Any,2}) = a + (-b)
```
devmotion, line 214: Similarly here, `-(::AbstractFill{<:Any,2}, ::UniformScaling)` falls back to `+(::AbstractFill{<:Any,2}, ::UniformScaling)`: https://github.com/JuliaLang/julia/blob/e3b681c6872d87d74db63dd7ec3da2beee8e3ae4/stdlib/LinearAlgebra/src/uniformscaling.jl#L159

src/fillalgebra.jl
```diff
@@ -261,6 +261,33 @@ end
...
+
+function -(a::UniformScaling{T}, b::Zeros{V, 2}) where {T <: Number, V}
+    n = LinearAlgebra.checksquare(b)
+    return Diagonal(Fill{promote_type(T,V)}(a.λ, n))
+end
```
devmotion, line 276: In principle, this definition could be removed since the default definition in src/FillArrays.jl would be `-(a::UniformScaling, b::Zeros) = a + (-b)` which by default (in LinearAlgebra) ends up calling `(-b) + a` which is covered by the definition for `+` above. However, adding this definition seemed more direct, clearer, and simpler to reason about.

src/FillArrays.jl
dlfivefifty: Not every `AbstractFill` can be converted to an `Array`, e.g. infinite arrays:
```julia
julia> using InfiniteArrays

julia> Fill(1,∞,∞) + Eye{Int}(∞)
(ℵ₀×ℵ₀ Fill{Int64, 2, Tuple{InfiniteArrays.OneToInf{Int64}, InfiniteArrays.OneToInf{Int64}}} with indices OneToInf()×OneToInf()) .+ (ℵ₀×ℵ₀ Diagonal{Int64, Ones{Int64, 1, Tuple{InfiniteArrays.OneToInf{Int64}}}} with indices OneToInf()×OneToInf()) with indices OneToInf()×OneToInf():
 2  1  1  1  1  1  1  1  1  1  1  1  1  …  
 1  2  1  1  1  1  1  1  1  1  1  1  1     
 1  1  2  1  1  1  1  1  1  1  1  1  1     
 1  1  1  2  1  1  1  1  1  1  1  1  1     
 1  1  1  1  2  1  1  1  1  1  1  1  1     
 1  1  1  1  1  2  1  1  1  1  1  1  1  …  
 1  1  1  1  1  1  2  1  1  1  1  1  1     
 1  1  1  1  1  1  1  2  1  1  1  1  1     
 1  1  1  1  1  1  1  1  2  1  1  1  1     
 1  1  1  1  1  1  1  1  1  2  1  1  1     
 1  1  1  1  1  1  1  1  1  1  2  1  1  …  
 1  1  1  1  1  1  1  1  1  1  1  2  1     
 1  1  1  1  1  1  1  1  1  1  1  1  2     
 1  1  1  1  1  1  1  1  1  1  1  1  1     
 1  1  1  1  1  1  1  1  1  1  1  1  1     
 1  1  1  1  1  1  1  1  1  1  1  1  1  …  
 1  1  1  1  1  1  1  1  1  1  1  1  1     
 1  1  1  1  1  1  1  1  1  1  1  1  1     
 ⋮              ⋮              ⋮        ⋱  
```
So:
```suggestion
+(a::AbstractFill{<:Any,2}, b::UniformScaling) = a + Diagonal(Fill(b.λ,(axes(a,1),)))
```

src/fillalgebra.jl
dlfivefifty: With my suggestion above the special case for `Zeros` is not needed.

dlfivefifty (changes requested):

src/FillArrays.jl
devmotion: Ah that's a good suggestion, I didn't consider arrays for which `Array` is not defined.

src/fillalgebra.jl
devmotion: I removed the definition for `Zeros`.

src/fillalgebra.jl
```diff
@@ -261,7 +261,6 @@ end
 -(a::AbstractRange{T}, b::Zeros{V, 1}) where {T, V} = a + b
```
devmotion, line 263:
```suggestion


```

devmotion:
@dlfivefifty I made the suggested simplifications and added some comments that explain why these methods are implemented. I also noticed that https://github.com/JuliaLang/julia/pull/28883 added definitions for `+(::Diagonal, ::UniformScaling)` etc. to base and hence in Julia >= 1.2.0 (was added in rc1) it is sufficient to define `+(::RectDiagonal, ::UniformScaling)` instead of `+(::RectOrDiagonal, ::UniformScaling)` etc. This also avoids the method ambiguity issues, and hence allowed to reduce the code even further. Clearly, this would require to bump the Julia compat of FillArrays to Julia >= 1.2.0 - would you be fine with this? Otherwise I could add back the definitions of `+(::Diagonal{T, <:AbstractFill{T, 1}}, ::UniformScaling)` only for Julia versions without https://github.com/JuliaLang/julia/pull/28883 (I don't think it makes sense to add them for newer versions).

dlfivefifty:
I think it's fine to raise the minimum version to v1.6 now that it's the LTR.

We should treat this as a breaking change anyways in case it introduces ambiguities elsewhere.

devmotion:
I updated the PR and removed support for Julia < 1.6. Would you like the deprecations to be removed in the breaking release?

dlfivefifty:
Ye

devmotion:
OK, I removed the deprecations (it seems only two constructors of `Eye` were deprecated).

dlfivefifty (approved):

devmotion:
Can this PR be merged and released? I would like to use the functionality in KernelFunctions :slightly_smiling_face:

devmotion:
Thank you!

KristofferC:
https://github.com/JuliaMatrices/ArrayLayouts.jl/blob/915da1f7952a754dd9371ea1b82e785c0fe70d92/Project.toml#L12 needs update.
https://github.com/JuliaArrays/BlockArrays.jl/blob/0d30a978c77778ae25a8907e4897bc853e87f52c/Project.toml#L12 needs update.

PR #172, opened 2022-03-07
Handle changed `copy_oftype`

rikhuijzer:
This fixes the tests on Julia 1.8-beta1. This fixes the underlying problem of https://github.com/JuliaArrays/LazyArrays.jl/pull/205:

For example, on Julia 1.8:
```julia
julia> Base.broadcasted(-, 1:5, Zeros{Int}(5))
5-element Vector{Int64}:
 1
 2
 3
 4
 5
```
which used to be
```julia
julia> Base.broadcasted(-, 1:5, Zeros{Int}(5))
1:5
```

This was caused in changes to `LinearAlgebra.copy_oftype` in https://github.com/JuliaLang/julia/pull/40831. As far as I understand, the meaning of `copy_oftype` used to be wrong and is now correct. I've solved the problem here by calling `copy` directly instead of `copy_oftype`.

### DIFF
--- a/src/fillbroadcast.jl
+++ b/src/fillbroadcast.jl
@@ -197,14 +197,11 @@ function broadcasted(::DefaultArrayStyle{1}, ::typeof(*), a::AbstractRange{V}, b
     return _range_convert(AbstractVector{promote_type(T,V)}, a)
 end
 
-copy_oftype(A::AbstractArray{T}, ::Type{T}) where {T} = copy(A)
-copy_oftype(A::AbstractArray{T,N}, ::Type{S}) where {T,N,S} = convert(AbstractArray{S,N}, A)
-
 for op in (:+, -)
     @eval begin
         function broadcasted(::DefaultArrayStyle{1}, ::typeof($op), a::AbstractVector{T}, b::Zeros{V,1}) where {T,V}
             broadcast_shape(axes(a), axes(b)) == axes(a) || throw(ArgumentError("Cannot broadcast $a and $b. Convert $b to a Vector first."))
-            copy_oftype(a, promote_type(T,V))
+            map(promote_type(T, V), a)
         end
 
         broadcasted(::DefaultArrayStyle{1}, ::typeof($op), a::AbstractFill{T,1}, b::Zeros{V,1}) where {T,V} =
@@ -214,7 +211,7 @@ end
 
 function broadcasted(::DefaultArrayStyle{1}, ::typeof(+), a::Zeros{T,1}, b::AbstractVector{V}) where {T,V}
     broadcast_shape(axes(a), axes(b))
-    copy_oftype(b, promote_type(T,V))
+    map(promote_type(T, V), b)
 end
 
 broadcasted(::DefaultArrayStyle{1}, ::typeof(+), a::Zeros{V,1}, b::AbstractFill{T,1}) where {T,V} = 
--- a/src/fillbroadcast.jl
+++ b/src/fillbroadcast.jl
@@ -197,11 +197,16 @@ function broadcasted(::DefaultArrayStyle{1}, ::typeof(*), a::AbstractRange{V}, b
     return _range_convert(AbstractVector{promote_type(T,V)}, a)
 end
 
+_copy_oftype(A::AbstractArray{T,N}, ::Type{T}) where {T,N} = copy(A)
+_copy_oftype(A::AbstractArray{T,N}, ::Type{S}) where {T,N,S} = convert(AbstractArray{S,N}, A)
+_copy_oftype(A::AbstractRange{T}, ::Type{T}) where T = copy(A)
+_copy_oftype(A::AbstractRange{T}, ::Type{S}) where {T,S} = map(S, A)
+
 for op in (:+, -)
     @eval begin
         function broadcasted(::DefaultArrayStyle{1}, ::typeof($op), a::AbstractVector{T}, b::Zeros{V,1}) where {T,V}
             broadcast_shape(axes(a), axes(b)) == axes(a) || throw(ArgumentError("Cannot broadcast $a and $b. Convert $b to a Vector first."))
-            map(promote_type(T, V), a)
+            _copy_oftype(a, promote_type(T,V))
         end
 
         broadcasted(::DefaultArrayStyle{1}, ::typeof($op), a::AbstractFill{T,1}, b::Zeros{V,1}) where {T,V} =
@@ -211,7 +216,7 @@ end
 
 function broadcasted(::DefaultArrayStyle{1}, ::typeof(+), a::Zeros{T,1}, b::AbstractVector{V}) where {T,V}
     broadcast_shape(axes(a), axes(b))
-    map(promote_type(T, V), b)
+    _copy_oftype(b, promote_type(T,V))
 end
 
 broadcasted(::DefaultArrayStyle{1}, ::typeof(+), a::Zeros{V,1}, b::AbstractFill{T,1}) where {T,V} = 
--- a/src/fillbroadcast.jl
+++ b/src/fillbroadcast.jl
@@ -197,21 +197,24 @@ function broadcasted(::DefaultArrayStyle{1}, ::typeof(*), a::AbstractRange{V}, b
     return _range_convert(AbstractVector{promote_type(T,V)}, a)
 end
 
+copy_oftype(A::AbstractArray{T}, ::Type{T}) where {T} = copy(A)
+copy_oftype(A::AbstractArray{T,N}, ::Type{S}) where {T,N,S} = convert(AbstractArray{S,N}, A)
+
 for op in (:+, -)
     @eval begin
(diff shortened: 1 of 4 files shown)


rikhuijzer:
This PR assumes that the following behavior is correct:

```julia
julia> LinearAlgebra.copy_oftype(1:5, Int)
5-element Vector{Int64}:
 1
 2
 3
 4
 5
```
versus the old behavior (Julia 1.7.2)
```julia
julia> LinearAlgebra.copy_oftype(1:5, Int)
1:5
```

rikhuijzer:
To verify this PR, I've ran the tests at https://github.com/JuliaArrays/LazyArrays.jl/pull/205 with this PR. On Julia 1.7.2 all tests pass and on Julia 1.8-beta1, only 2 tests fail. Those 2 failing tests are due to problems with `copy_oftype` in FillArrays. I'm working on a PR for that too.

@dlfivefifty What do you think?

dlfivefifty:
No this won't due as we need to promote the type.

I think just copy over the old `copy_oftype`.

Really LinearAlgebra should call it `mutablecopy_oftype` as `copy` is not in general mutable.

daanhb:
Hmm, `copy_oftype` is internal to LinearAlgebra and it is somewhat surprising that you would need it here. If `copy` works here then I guess that's fine. If you want to preserve structure and mutability, but possibly change the eltype, then `convert(AbstractArray{T}, myarray)` may do the trick. That is what the old `copy_oftype` did in most cases. Or `AbstractArray{T}(myarray)`.

rikhuijzer:
> Hmm, `copy_oftype` is internal to LinearAlgebra and it is somewhat surprising that you would need it here. If `copy` works here then I guess that's fine. If you want to preserve structure and mutability, but possibly change the eltype, then `convert(AbstractArray{T}, myarray)` may do the trick. That is what the old `copy_oftype` did in most cases. Or `AbstractArray{T}(myarray)`.

In this case, I copied the `copy` from the method that was called in Julia 1.7.2 so maybe there the generality of `convert(AbstractArray{T}, myarray)` or `AbstractArray{T}(myarray)` isn't needed. In any case, I suspect that `copy` will be the quickest.

But well, you guys know more so just let me know what you prefer :+1:

rikhuijzer:
> No this won't due as we need to promote the type.
> 
> I think just copy over the old `copy_oftype`.
> 
> Really LinearAlgebra should call it `mutablecopy_oftype` as `copy` is not in general mutable.

That's what I did. I copied the content of the method that was called.

daanhb:
> In this case, I copied the `copy` from the method that was called in Julia 1.7.2 so maybe there the generality of `convert(AbstractArray{T}, myarray)` or `AbstractArray{T}(myarray)` isn't needed. In any case, I suspect that `copy` will be the quickest.
> 
> But well, you guys know more so just let me know what you prefer 👍

As far as I'm concerned, whatever works :-) I'm just making suggestions to resolve any issues here caused by changes to `copy_oftype`...

rikhuijzer:
With `copy`:

```julia
julia> (1:5) .+ Zeros(5)
1:5
```

This is the correct output here.

With  `AbstractArray{promote_type(T, V)}(a)` and `convert(AbstractArray{promote_type(T, V)}, a)`:

```julia
julia> (1:5) .+ Zeros(5)
5-element Vector{Float64}:
 1.0
 2.0
 3.0
 4.0
 5.0
```

which is the original problem and caused by

```julia
julia> UnitRange{Int} <: AbstractVector{Int}
true

julia> Vector{Int} <: AbstractVector{Int}
true
```

The method for Julia 1.7 for `LinearAlgebra.copy_oftype(1:5, Int)` is `copy`:

```julia
copy_oftype(A::AbstractArray{T}, ::Type{T}) where {T} = copy(A)
```

[source](https://github.com/JuliaLang/julia/blob/bf534986350a991e4a1b29126de0342ffd76205e/stdlib/LinearAlgebra/src/LinearAlgebra.jl#L358)

If there is some case where `copy` isn't suitable, it would be nice if we could come with such a counterexample and add a test for it

dlfivefifty:
The correct output for that example is ` 1.0:5.0 ` since the type needs to be promoted

daanhb:
Hmm, please don't solve it like this. (You are now copying an old internal function from LinearAlgebra with unclear semantics, without changing its name, while the function in LinearAlgebra has meanwhile changed.)

There is an underlying problem and it may be better to fix that. In your example, the problem is, I think, not with FillArrays or with `copy_oftype`, the problem is:
```julia
julia> convert(AbstractVector{Float64}, 1:5)
5-element Vector{Float64}:
 1.0
 2.0
 3.0
 4.0
 5.0
```
This is a result I would disagree with in principle, but some code out there relies on this conversion returning a mutable array. So it would be a breaking change to change it in Base and I didn't even raise the issue. (I did check for occurrences of `copy_oftype` in the wild also before changing it and there were very few, you are unlucky here!)

The choice in Base so far has been to support the syntax `map(T, myarray)` instead:
```julia
julia> map(Float64, 1:5)
1.0:1.0:5.0
```
Could you experiment with that? I don't think it will work for all structured matrices though, as they all have to support this syntax and I don't think they all do. On the other hand, `convert(AbstractArray{T}, A)` should work for all structured matrices, but not for ranges. Perhaps it is sufficient to special case ranges in your code, and use `convert(AbstractArray{T}, A)` otherwise?

Sorry for the hassle!

daanhb:
Ideally, you no longer use `copy_oftype` anymore, and in the future the function will hopefully be gone from LinearAlgebra. It is only useful in certain cases to obtain a structured array with a different element type which is suitable for in-place factorization algorithms. It works for now, but there is nothing generic about it.

rikhuijzer:
`map` works great

test/runtests.jl
```diff
@@ -1053,8 +1053,6 @@ end
 
     @test copy(m) ≡ m
     @test copy(D) ≡ D
-    @test LinearAlgebra.copy_oftype(m, Int) ≡ Eye{Int}(10)
-    @test LinearAlgebra.copy_oftype(D, Float64) ≡ Diagonal(Fill(2.0,10))
```
rikhuijzer, line 1057: I'm not sure what this tests. It was added as part of https://github.com/JuliaArrays/FillArrays.jl/pull/142. The `copy` for `Eye` and `Diagonal` still works, so I think it's safe to remove these two

src/fillbroadcast.jl
```diff
@@ -201,17 +201,17 @@ for op in (:+, -)
     @eval begin
         function broadcasted(::DefaultArrayStyle{1}, ::typeof($op), a::AbstractVector{T}, b::Zeros{V,1}) where {T,V}
             broadcast_shape(axes(a), axes(b)) == axes(a) || throw(ArgumentError("Cannot broadcast $a and $b. Convert $b to a Vector first."))
-            LinearAlgebra.copy_oftype(a, promote_type(T,V))
+            map(promote_type(T, V), a)
```
dlfivefifty, line 204: Note their's a minor performance degradation:
```julia
julia> @btime copy(a);
  734.641 μs (2 allocations: 7.63 MiB)

julia> @btime map(Float64,a);
  826.656 μs (3 allocations: 7.63 MiB)
```

dlfivefifty (changes requested):
I'm worried this is a bit too much of a "pun". In particular, it requires types to overload `map` and so won't work if only `copy` is overloaded.

This is my proposal, add the following function and call it:
```julia
_copy_oftype(A::AbstractArray{T,N}, ::Type{T}) where {T,N} = copy(A)
_copy_oftype(A::AbstractArray{T,N}, ::Type{S}) where {T,N,S} = convert(AbstractArray{S,N}, A)
_copy_oftype(A::AbstractRange{T}, ::Type{T}) where T = copy(A)
_copy_oftype(A::AbstractRange{T}, ::Type{S}) where {T,S} = map(S, A)
```

dlfivefifty:
> without changing its name, while the function in LinearAlgebra has meanwhile changed

I think the problem is the name of the new `LinearAlgebra.copy_oftype` should be renamed to `LinearALgebra.copymutable_oftype`, since `copy` on its own does not imply mutable. I'm planning to make a PR.

daanhb:
> > without changing its name, while the function in LinearAlgebra has meanwhile changed
> 
> I think the problem is the name of the new `LinearAlgebra.copy_oftype` should be renamed to `LinearALgebra.copymutable_oftype`, since `copy` on its own does not imply mutable. I'm planning to make a PR.

Fair enough I think. It is indeed intended to be mutable.

dlfivefifty (approved):

rikhuijzer:
By the way, can we use this same fix for ArrayLayouts.jl? Then, I'll make a PR there too

dlfivefifty:
Probably. I would just call `FillArrays._copy_oftype` from there (I know relying on internals is not generally good, but in this case they are both my packages)

daanhb:
Is it correct that it really just comes down to the behaviour of ranges after conversion, or am I missing something? Is there another example where the new `copy_oftype` does not do what you want here?
I'm fine with any solution here, I'm just wondering about possible side-effects of `copy_oftype` elsewhere that we're not aware of.

Issue #161, opened 2021-09-27
Support setindex


dlfivefifty:
The following should be a fast way of creating `e_k` (the vector of all zeros but with 1 in the k-th entry):
```julia
julia> using FillArrays

julia> Base.setindex(Zeros(5), 2, 3)
ERROR: MethodError: no method matching setindex(::Zeros{Float64, 1, Tuple{Base.OneTo{Int64}}}, ::Int64, ::Int64)
Closest candidates are:
  setindex(::Tuple, ::Any, ::Integer) at tuple.jl:49
  setindex(::CartesianIndex, ::Any, ::Any) at multidimensional.jl:100
  setindex(::NamedTuple, ::Any, ::Symbol) at namedtuple.jl:355
Stacktrace:
 [1] top-level scope
   @ REPL[2]:1
```

mcabbott:
What are you picturing this returning? 

Zygote's `OneElement` is something very close to this, which could perhaps be moved upstream if it has other uses. Allows any `ndims`, any axes.

Flux also has a `OneHotVector` which is more restricted, stores just a `UInt32`. A vector of these can be re-interpreted as integers. There was talk of moving that out, too, but perhaps not a good fit here.

dlfivefifty:
Yeah, I guess the proposal would be to move `OneElement` here

PR #235, opened 2023-03-28
Move OneElement from Zygote and overload setindex (#161)

dlfivefifty:
@mcabbott FYI.

### DIFF
--- a/src/oneelement.jl
+++ b/src/oneelement.jl
@@ -1,0 +1,16 @@
+"""
+    OneElement(val, ind, axes) <: AbstractArray
+Extremely simple `struct` used for the gradient of scalar `getindex`.
+"""
+struct OneElement{T,N,I,A} <: AbstractArray{T,N}
+  val::T
+  ind::I
+  axes::A
+  OneElement(val::T, ind::I, axes::A) where {T<:Number, I<:NTuple{N,Int}, A<:NTuple{N,AbstractUnitRange}} where {N} = new{T,N,I,A}(val, ind, axes)
+end
+
+# OneElement(val, inds::Int...) = OneElement(val, inds)
+
+Base.size(A::OneElement) = map(length, A.axes)
+Base.axes(A::OneElement) = A.axes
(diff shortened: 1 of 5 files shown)


src/oneelement.jl
```diff
@@ -0,0 +1,27 @@
...
+end
+
+OneElement(val, inds::NTuple{N,Int}, sz::NTuple{N,Integer}) where N = OneElement(val, inds, oneto.(sz))
+OneElement(val, inds::Int, sz::Int) = OneElement(val, (inds,), (sz,))
+OneElement(inds::Int, sz::Int) = OneElement(1, inds, sz)
+OneElement{T}(inds::Int, sz::Int) where T = OneElement(one(T), inds, sz)
+
+Base.size(A::OneElement) = map(length, A.axes)
+Base.axes(A::OneElement) = A.axes
+Base.getindex(A::OneElement{T,N}, i::Vararg{Int,N}) where {T,N} = ifelse(i==A.ind, A.val, zero(T))
```
mcabbott, line 19: Should this have boundschecks? I didn't bother for Zygote, but perhaps for wider use... 

Fill does:
```julia
julia> Fill(1,2,3)
2×3 Fill{Int64}, with entries equal to 1

julia> ans[3,4]
ERROR: BoundsError: attempt to access 2×3 Fill{Int64} at index [3, 4]
```

src/oneelement.jl
```diff
@@ -0,0 +1,27 @@
+"""
+    OneElement(val, ind, axes) <: AbstractArray
+Extremely simple `struct` used for the gradient of scalar `getindex`.
```
mcabbott, line 3: It's possible that this should have friendlier constructors, which e.g. call `to_indices` to convert any other index input, and allow you to supply size not axes? 

Or it could just mention that `setindex(Zeros(), ...)` is the intended interface.

dlfivefifty: This was already added, I've added more docs.

src/oneelement.jl
dlfivefifty: fixed!

PR #352, opened 2024-03-14
using EltypeExtensions: elconvert

putianyi889:
The goal is to let multiple packages (FillArrays, QuasiArrays, LazyArrays, etc.) use the same `elconvert` while not having to depend on each other.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -22,7 +22,7 @@ FillArraysStatisticsExt = "Statistics"
 [compat]
 Aqua = "0.8"
 Documenter = "1"
-EltypeExtensions = "0.1.0"
+EltypeExtensions = "0.1.1"
 Infinities = "0.1"
 LinearAlgebra = "1.6"
 PDMats = "0.11.17"


dlfivefifty:
Can you add a description to ElTypeExtensions.jl README?

putianyi889:
> Can you add a description to ElTypeExtensions.jl README?

done.

putianyi889:
@dlfivefifty can you merge this?

dlfivefifty (approved):

dlfivefifty:
I think this is fine but perhaps we should move EltypeExtensions.jl to JuliaArrays first?

It's better for it to be a community developed pacakge.

dlfivefifty:
I think `elconvert` is not a great name. `convert_eltype` would be preferable.

putianyi889:
> It's better for it to be a community developed pacakge.

This is true.

> move EltypeExtensions.jl to JuliaArrays

but this solution doesn't seem great. EltypeExtensions supports all kinds of collections, for example, it could be used for DomainSets.

putianyi889:
> I think `elconvert` is not a great name. `convert_eltype` would be preferable.

Updated to EltypeExtensions@0.1

PR #422, opened 2025-12-05
Zeros and OneElement addition/subtraction

max-vassili3v:
This pull request alters some behaviour of addition/subtraction of the `Zeros` and `OneElement` types. It ensures that the sparsity of any array structure that undergoes addition or subtraction with one of these types preserves its sparsity by avoiding the materialisation of dense arrays where possible:

-Adding or subtracting a `Zeros` and any array `v` now preserves the type of `v` (while still obeying type promotion rules)
-Adding or subtracting a `OneElement` to any array `v` is now O(1) and preserves the type of `v` unless the underlying `setindex!` operation is invalid (e.g adding a `OneElement` with a nondiagonal entry to a `Diagonal`), in which case a dense array is materialised
-Adding or subtracting two `OneElements` now returns a `OneElement` if their indices are the same. Otherwise it materialises a dense vector. If `SparseArrays` is imported, it instead creates a SparseArray.

Changes have been covered with unit tests

PR #427, opened 2026-01-06
Add Static Arrays Extension

max-vassili3v:
This is a continuation of [this PR](https://github.com/JuliaArrays/FillArrays.jl/pull/422/files).

This PR adds an extension of StaticArrays.jl. This ends up fixing an issue where adding or subtracting by `Zeros` always returns a dense `Array `. By adding this extension, we resolve ambiguity issues and thus are able to override `+(a::AbstractArray, b::Zeros)` so that it preserves the type of `a`.


dlfivefifty (approved):

dlfivefifty:
Can you bump the minor version? (So v1.16)

PR #442, opened 2026-07-19
remove dependancy on internal base func

rokke-git:
it doesn't seem like it even changes anything, just calls an expensive `invoke` for some reason? I don't know what this was trying to do but I'll fix it if it breaks anything somehow; trying to clean up these internal array print functions in base and found your package complaining that "`print_matrix_row` not defined in `Base`" when importing a downstream package

edit: ok, I think I'm starting to see what this is trying to do. prob should have started at block arrays as mentioned in the comment...

### DIFF
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -725,23 +725,6 @@ include("trues.jl")
 Base.replace_in_print_matrix(::AbstractZeros, ::Integer, ::Integer, s::AbstractString) =
     Base.replace_with_centered_mark(s)
 
-# following support blocked fill array printing via
-# BlockArrays.jl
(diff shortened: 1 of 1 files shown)


dlfivefifty:
But coverage is 100%? How is it that it can't be called?

As the comment makes clear, this is to support downstream BlockArrays.jl

rokke-git:
sorry, like I was saying in my edit; I was going too fast. it does get called; but when looking at code only in your repo it just sends execution back down the normal display pipeline. this section should maybe be in `ext`, and `axes_print_matrix_row` should maybe be explicitly imported from `BlockArrays`, but I think it works as is. I'm looking at the block arrays side of this first, that's why I marked as draft to come back here after I look at that side

wow, just realized you can run the origin's ci workflows from pull requests on your own fork...

PR #443, opened 2026-07-23
Remove method overloads that duplicate Base/LinearAlgebra generics

JamesWrigley:
These all have generic definitions upstream that do the same thing. Using Base's unary `-` also fixes a wee bug by preserving the axes of the inputs such that `-Fill(2, (1:3,)) == Fill(-2, (1:3,))`. The old implementation used `size()` so the result would be `-Fill(2, (3,))`.

Discovered with help from Claude :robot:

### DIFF
--- a/src/FillArrays.jl
+++ b/src/FillArrays.jl
@@ -5,7 +5,7 @@ using LinearAlgebra
 import Base: size, getindex, setindex!, IndexStyle, checkbounds, convert,
     +, -, *, /, \, diff, sum, cumsum, maximum, minimum, sort, sort!,
     any, all, axes, isone, iszero, iterate, unique, allunique, permutedims, inv,
-    copy, vec, setindex!, count, ==, reshape, map, zero,
+    copy, setindex!, count, ==, reshape, map, zero,
     show, view, in, mapreduce, one, reverse, promote_op, promote_rule, repeat,
     parent, similar, issorted, add_sum, mul_prod, accumulate, OneTo, permutedims
 
@@ -512,7 +512,7 @@ function diag(E::Eye, k::Integer=0)
 end
 
 # These should actually be in StdLib, LinearAlgebra.jl, for all Diagonal
-for f in (:permutedims, :triu, :triu!, :tril, :tril!, :copy)
+for f in (:triu, :triu!, :tril, :tril!, :copy)
     @eval ($f)(IM::Diagonal{<:Any,<:AbstractFill}) = IM
 end
 
--- a/src/fillalgebra.jl
+++ b/src/fillalgebra.jl
@@ -1,7 +1,3 @@
-## vec
-
-vec(a::AbstractFill) = fillsimilar(a, length(a))
-
 ## Transpose/Adjoint
 # cannot do this for vectors since that would destroy scalar dot product
 
@@ -426,7 +422,6 @@ end
 # Addition and Subtraction
 +(a::AbstractFill) = a
 -(a::AbstractZeros) = a
--(a::AbstractFill) = Fill(-getindex_value(a), size(a))
 
 # special-cased for type-stability, as Ones + Ones is not a Ones
 Base.reduce_first(::typeof(+), x::AbstractOnes) = Fill(Base.reduce_first(+, getindex_value(x)), axes(x))


jishnub:
Some of these are for performance and dead-code elimination, so it's worth checking that

dlfivefifty:
I will be surprised if the downstream tests pass

JamesWrigley:
CI looks good. Do you know which methods are there for performance? `isassigned()` and `count()` are the ones that jump out at me.

src/FillArrays.jl
```diff
@@ -34,11 +34,6 @@ const AbstractFillVecOrMat{T} = Union{AbstractFillVector{T},AbstractFillMatrix{T
 
 ==(a::AbstractFill, b::AbstractFill) = axes(a) == axes(b) && getindex_value(a) == getindex_value(b)
 
-@inline function Base.isassigned(F::AbstractFill, i::Integer...)
-    @boundscheck checkbounds(Bool, F, to_indices(F, i)...) || return false
-    return true
-end
```
dlfivefifty, line 40: The default `isassigned` uses a slow try-catch block. Please restore these.

src/FillArrays.jl
```diff
@@ -694,11 +689,6 @@ function all(f::Function, x::AbstractFill)
     fval = f(getindex_value(x))
     return all((fval,))
 end
-any(x::AbstractFill) = any(identity, x)
```
dlfivefifty, line 697: `any`/`all` will iterate over all the entries. Please restore these

src/FillArrays.jl
```diff
@@ -694,11 +689,6 @@ function all(f::Function, x::AbstractFill)
...
-all(x::AbstractFill) = all(identity, x)
-
-count(x::AbstractOnes{Bool}) = length(x)
```
dlfivefifty, line 700: ditto

dlfivefifty (changes requested):

dlfivefifty:
Most of these overloads are very clearly needed to avoid looping over all entries.

src/FillArrays.jl
JamesWrigley: I'm not sure I follow, don't these also iterate over all the entries and are the same definitions as in Base? https://github.com/JuliaLang/julia/blob/0fe418d406adbaea5d9b28a5a8410f9e2522d028/base/reducedim.jl#L991

dlfivefifty: Why would they???? The point of fill arrays is you know all the entries:

https://github.com/JuliaArrays/FillArrays.jl/blob/7b96cc2496edeae45649883d246138461e16b903/src/FillArrays.jl#L685

JamesWrigley: Ah I'm blind, didn't see the `any(::Function, x)` definitions above it :facepalm:

src/FillArrays.jl
JamesWrigley: Fixed in 23f248c.

src/FillArrays.jl
JamesWrigley: Fixed in 23f248c.

src/FillArrays.jl
JamesWrigley: Fixed in 23f248c.

JamesWrigley:
I also restored the `count(::AbstractZeros)` method since that's an optimization too. So now the remaining deleted methods are:
- `vec()` (definition in Base calls `reshape()` internally which ends up calling `fillsimilar()` too)
- `permutedims()` (upstream definition does the same thing, and we didn't have a two-arg definition anyway)
- Unary `-`, which fixes the multidimensional case and calls `broadcastable` internally.

dlfivefifty (approved):

src/FillArrays.jl
JamesWrigley: Hmm on second thoughts aren't these ok to be deleted since both they and the Base versions will end up calling `any(::Function, ::AbstractFill)`?

dlfivefifty: One beautiful thing about julia is you can easily see what is happening, using `@which` or the debugger. The answer: no, Base does not call `any(identity, ::AbstractFill)` but instead calls `_any`:
```julia
[1/2] any(a; dims) at /Users/solver/.julia/juliaup/julia-1.12.6+0.aarch64.apple.darwin14/Julia-1.12.app/Contents/Resources/julia/share/julia/base/reducedim.jl:988
>988  any(a::AbstractArray; dims=:)              = _any(a, dims)
```
This eventually iterates over a loop:
```julia
[1/4] _any(f, itr, #unused#) at /Users/solver/.julia/juliaup/julia-1.12.6+0.aarch64.apple.darwin14/Julia-1.12.app/Contents/Resources/julia/share/julia/base/anyall.jl:121
 120  @eval function _any(f, itr::$ItrT, ::Colon)
>121      $(ItrT === Tuple ? :(@_terminates_locally_meta) : :nothing)
 122      anymissing = false
 123      for x in itr
```

I appreciate your desire to remove code but please make sure you double check that that the default actually avoids iteration.

JamesWrigley: Ah right, I missed that.

PR #451, opened 2026-09-21
Revert "Update Random requirement from 1 to 1, < 0.0.1 in /docs"

dlfivefifty:
Reverts JuliaArrays/FillArrays.jl#450


dlfivefifty:
The "<v0.0.1" makes no sense to me

